PR #95, opened 2014-12-31
Test all regression models with basic test

az0:
On my system this test does not pass for many models, so I do not necessarily expect you to accept it as is.  However, what do you think of the general idea?  Where this test fails, is it the test or caret? 

At least these models pass
````R
models<-c('lm','earth','bagEarth', 'pls')
````


az0:
Now I am running Travis CI on my fork, so I can test without this pull request

PR #101, opened 2015-01-13
Example now preserves integrity of the data in when

jschadewald:
In caret 6.0-41, when running example(dummyVars), both steps that make use of the levels() function result in modifying the original data.

Before assigning levels, we have:
> when
       time day
1 afternoon Mon
2     night Mon
3 afternoon Mon
4   morning Wed
5   morning Wed
6   morning Fri
7   morning Sat
8 afternoon Sat
9 afternoon Fri

In the 6.0-41 release, setting levels changes the values like so:
> when
       time day
1   morning Tue
2     night Tue
3   morning Tue
4 afternoon Thu
5 afternoon Thu
6 afternoon Mon
7 afternoon Wed
8   morning Wed
9   morning Mon

The fix is to use named lists, rather than vectors, to set the levels as desired.

### DIFF
--- a/pkg/caret/man/dummyVars.Rd
+++ b/pkg/caret/man/dummyVars.Rd
@@ -128,8 +128,11 @@ when <- data.frame(time = c("afternoon", "night", "afternoon",
                            "Wed", "Wed", "Fri",
                            "Sat", "Sat", "Fri"))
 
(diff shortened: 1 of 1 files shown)


topepo:
Thanks for correcting that. It is very subtle.

PR #106, opened 2015-01-21
Trim code for glm and bayesglm

zachmayer:


zachmayer:
@topepo I'm seeing the following note in `R CMD CHECK`:
```{R}
* checking R code for possible problems ... NOTE
bagImp: no visible global function definition for ‘trim’
```
It seems to predate my PR.  Is it possible the `bagImp` still references the old `trim` S3 method?

topepo:
I just checked in some code that added the S3 `trim` class back in along with a method for `train`

PR #125, opened 2015-02-27
Call the generic function instead of the method

xrobin:
The upcoming pROC version 1.8 has a refactored NAMESPACE and won't allow direct calls to S3 methods. Instead, the generic function must be called. This fix is necessary to push fix xrobin/pROC#4 to CRAN.

Thanks!

### DIFF
--- a/pkg/caret/R/aaa.R
+++ b/pkg/caret/R/aaa.R
@@ -220,7 +220,7 @@ twoClassSummary <- function (data, lev = NULL, model = NULL)
   requireNamespaceQuietStop('pROC')
   if (!all(levels(data[, "pred"]) == levels(data[, "obs"]))) 
     stop("levels of observed and predicted data do not match")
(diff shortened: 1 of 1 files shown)


zachmayer:
:+1: From me, if tests pass.

xrobin:
Thanks! Unfortunately it looks like a test failed...

    Your test run exceeded 50 minutes.

Not sure if we're looking at a regression? There's one more dispatch call, but that can't be significant compared to the computations in the roc() function...

zachmayer:
Yeah, I don't think that's you.  I'm going to take the vignette building out of the test code, that alone takes ~30 mins.



—

topepo:
Thanks for helping out.

xrobin:
Thanks, the tests with the new pROC version seem to pass now. Any idea when that fix will be available on CRAN?

topepo:
I'll shoot for a new release in a 2-3 weeks.

PR #127, opened 2015-03-15
Allow construction of RBF models for regression.

dratewka:
After adding support for regression type models to RBF networks everything seems to be working fine.

### DIFF
--- a/models/files/rbf.R
+++ b/models/files/rbf.R
@@ -1,7 +1,7 @@
 modelInfo <- list(label = "Radial Basis Function Network",
                   library = "RSNNS",
                   loop = NULL,
-                  type = c('Classification'),
+                  type = c('Classification','Regression'),
                   parameters = data.frame(parameter = c('size'),
                                           class = c('numeric'),
                                           label = c('#Hidden Units')),


zachmayer:
Looks good to me.  Have you tested that this works?

dratewka:
Yep, seems to work:

  > devtools::install_github('dratewka/caret/pkg/caret')
  > x <- matrix(seq(-10,10, length=100), ncol=1)
  > y <- sin(x) + rnorm(length(x))
  > m <- caret::train(x,y, method='rbfDDA')
  > mean((y-predict(m,x))^2)
  [1] 1.508441

zachmayer:
@topepo I think you can go ahead and merge this, it looks like a pretty straightforward change.

topepo:
This is relatively untested but it might work:

```r
mnLogLoss <- function(data, lev = NULL, model = NULL){
  eps <- 1e-15
  probs <- as.matrix(data[, lev, drop = FALSE])
  probs[probs > 1 - eps] <- 1 - eps
  probs[probs < eps] <- eps
  inds <- caret:::class2ind(data$obs)
  -mean(apply(inds*log(probs), 1, sum), na.rm = TRUE)
}
```

PR #138, opened 2015-04-07
Fix a couple of typos in manual pages.

ilarischeinin:


topepo:
Thanks for fixing those.

PR #156, opened 2015-05-25
Fix incorrect sampling in bagged avNNet.

timcdlucas:
avNNet with bag = TRUE currently samples without replacement. This therefore just reorders the rows.
avNNet now samples WITH replacement therefore giving proper bagging.

I don't think the same error occurs in other models but I'm not sure.


topepo:
Thanks. That is definitely a bug.

PR #168, opened 2015-06-15
vectorized implementation of findCorrelation

EdwinTh:
Function is implemented vectorized instead of using for loops, which speeds up the function remarkably.

Furthermore may we kindly suggest to omit the symmetry check, or make it optional. This causes memory problems on large feature sets and it's clearly stated that x should be a correlation matrix.

### DIFF
--- a/pkg/caret/R/findCorrelation.R
+++ b/pkg/caret/R/findCorrelation.R
@@ -1,15 +1,23 @@
-findCorrelation <- function(x, cutoff = .90){
+findCorrelation <- function(x, cutoff = .90, verbose = FALSE){
   averageCorr <- colMeans(abs(x))
   averageCorr <- as.numeric(as.factor(averageCorr))
   x[lower.tri(x, diag = TRUE)] <- NA
   combsAboveCutoff <- which(abs(x) > cutoff)
-  
+
   colsToCheck <- ceiling(combsAboveCutoff / nrow(x))
   rowsToCheck <- combsAboveCutoff %% nrow(x)
-  
+
   colsToDiscard <- averageCorr[colsToCheck] > averageCorr[rowsToCheck]
   rowsToDiscard <- !colsToDiscard
-  
+
+  if(verbose){
+    colsFlagged <- pmin(ifelse(colsToDiscard, colsToCheck, NA),
+                        ifelse(rowsToDiscard, rowsToCheck, NA), na.rm = TRUE)
+    cat('\n',paste('Combination row', rowsToCheck, 'and clolumn', colsToCheck,
+                   'is above the cut-off\n \t Flagging column', colsFlagged, '\n'
+                   ))
+  }
+
   deletecol <- c(colsToCheck[colsToDiscard], rowsToCheck[rowsToDiscard])
   deletecol <- unique(deletecol)
   deletecol
--- a/pkg/caret/R/findCorrelation.R
+++ b/pkg/caret/R/findCorrelation.R
@@ -1,45 +1,16 @@
-findCorrelation <- function(x, cutoff = 0.90, verbose = FALSE)
-{
-  varnum <- dim(x)[1]
+findCorrelation <- function(x, cutoff = .90){
+  averageCorr <- colMeans(abs(x))
+  averageCorr <- as.numeric(as.factor(averageCorr))
+  x[lower.tri(x, diag = TRUE)] <- NA
+  combsAboveCutoff <- which(abs(x) > cutoff)
   
-  if (!isTRUE(all.equal(x, t(x)))) stop("correlation matrix is not symmetric")
-  if (varnum == 1) stop("only one variable given")
+  colsToCheck <- ceiling(combsAboveCutoff / nrow(x))
+  rowsToCheck <- combsAboveCutoff %% nrow(x)
   
-  x <- abs(x)
+  colsToDiscard <- averageCorr[colsToCheck] > averageCorr[rowsToCheck]
+  rowsToDiscard <- !colsToDiscard
   
-  # re-ordered columns based on max absolute correlation
-  originalOrder <- 1:varnum
-  
-  averageCorr <- function(x) mean(x, na.rm = TRUE)
-  tmp <- x
-  diag(tmp) <- NA
-  
-  maxAbsCorOrder <- order(apply(tmp, 2, averageCorr), decreasing = TRUE)
-  x <- x[maxAbsCorOrder, maxAbsCorOrder]
-  newOrder <- originalOrder[maxAbsCorOrder]
-  
-  deletecol <- rep(F, varnum)
-  
-  for (i in 1:(varnum - 1)) {
-    if (deletecol[i]) next
-    for (j in (i + 1):varnum) {
-      if (!deletecol[i] & !deletecol[j]) {
-        if(verbose)
-          cat("Considering row\t", newOrder[i], 
-              "column\t", newOrder[j], 
-              "value\t", round(x[i,j], 3), "\n")  
-        if (x[i, j] > cutoff) {
-          if (mean(x[i, -i]) > mean(x[-j, j])) {
-            deletecol[i] <- T
-            if (verbose) cat("  Flagging column\t", newOrder[i], "\n")
-          }
-          else {
-            deletecol[j] <- T
-            if (verbose) cat("  Flagging column\t", newOrder[j], "\n")
-          }
-        }
-      }
-    }
-  }
-  newOrder[which(deletecol)]
+  deletecol <- c(colsToCheck[colsToDiscard], rowsToCheck[rowsToDiscard])
+  deletecol <- unique(deletecol)
+  deletecol
 }
--- a/pkg/caret/R/findCorrelation.R
+++ b/pkg/caret/R/findCorrelation.R
@@ -13,8 +13,10 @@ findCorrelation <- function(x, cutoff = .90, verbose = FALSE){
   if(verbose){
     colsFlagged <- pmin(ifelse(colsToDiscard, colsToCheck, NA),
                         ifelse(rowsToDiscard, rowsToCheck, NA), na.rm = TRUE)
+    values <- round(x[combsAboveCutoff], 3)
     cat('\n',paste('Combination row', rowsToCheck, 'and clolumn', colsToCheck,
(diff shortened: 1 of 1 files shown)


zachmayer:
You need to add a `verbose=TRUE` argument to the new function:
```{R}
Error in findCorrelation(corrMatrix, cutoff = 0.65, verbose = TRUE) : 
  unused argument (verbose = TRUE)
Execution halted
```

Or remove it from the examples for findCorrelation.

In the future, load the RStudio project and run `devtools::check()` prior to making a PR =D

EdwinTh:
Thanks, indeed forgot about that one. Will fix it right away and will look
into the devtools::check()

EdwinTh:
Verbose option added. Ran devtools::check(), this gave two warnings which did not seem to be due to the findCorrelation() function.

Since the function is vectorized, verbose = TRUE will only print the combinations with a correlation  of which the absolute value is greater than the cutoff.

topepo:
Edwin,

Thanks for the addition. This is much faster although it doesn't quite do the same thing. The current version does the calculations stage-wise (why the `for` loop was used). Theoretically the current version can give you a smaller subset while keeping the correlations to the same level as the new version. Here is an example from `?findCorrelation`:

```r
findCorrelation_new <- function(x, cutoff = .90){
  averageCorr <- colMeans(abs(x))
  averageCorr <- as.numeric(as.factor(averageCorr))
  x[lower.tri(x, diag = TRUE)] <- NA
  combsAboveCutoff <- which(abs(x) > cutoff)
  
  colsToCheck <- ceiling(combsAboveCutoff / nrow(x))
  rowsToCheck <- combsAboveCutoff %% nrow(x)
  
  colsToDiscard <- averageCorr[colsToCheck] > averageCorr[rowsToCheck]
  rowsToDiscard <- !colsToDiscard
  
  deletecol <- c(colsToCheck[colsToDiscard], rowsToCheck[rowsToDiscard])
  deletecol <- unique(deletecol)
  deletecol
}

corrMatrix <- diag(rep(1, 5))
corrMatrix[2, 3] <- corrMatrix[3, 2] <- .7
corrMatrix[5, 3] <- corrMatrix[3, 5] <- -.7
corrMatrix[4, 1] <- corrMatrix[1, 4] <- -.67
```

The results are different but have the same length:

```
> findCorrelation(corrMatrix, cutoff = .65)
[1] 3 4
> findCorrelation_new(corrMatrix, cutoff = .65)
[1] 3 1
```

I think this happens because `averageCorr` in your version uses all the columns whereas the version with the `for` loop uses only those that have not been eliminated at the current stage:

```
> colMeans(corrMatrix[-3,-3])
[1] 0.0825 0.2500 0.0825 0.2500
> colMeans(corrMatrix)[-3]
[1] 0.066 0.340 0.066 0.060
```

How much does this matter? I have no idea. I'm currently testing it on random matrices to see if the results differ much (see below; they don't seem to so far). 

I think that we should add an option to the current version (maybe call it `fast`) that uses your version and make a note that it may not give the same results as the original algorithm. 

It would be a good idea to write some unit tests for small examples like the one that I have above too.

Thanks,

Max

```r
## comparison code
for(i in 1:1000) {
  set.seed(i)
  ## make random correlation matrix
  R <- matrix(runif(10000), ncol=100) 
  R <- (R * lower.tri(R)) + t(R * lower.tri(R)) 
  diag(R) <- 1 
  ## test across different cutoffs
  for(j in seq(.1, .9, by = .1)) {
    old_results <- findCorrelation(R, cutoff = j)
    new_results <- findCorrelation_new(R, cutoff = j)
    
    if(length(old_results) != length(new_results)) {
      print(i, j)
    } else {
      if(!all(sort(length(old_results)) == sort(length(old_results))))
        print(i, j)
    }
  }
}
```

EdwinTh:
Fair point, providing both options is a good solution, however I would intuitively say that recalculating the mean correlations after each removal as in the original will yield slightly superior results. Maybe we can still improve the speed by only considering the combinations that are above the cutoff and that doesn't contain variables that are already discarded.

Regards,
Edwin

topepo:
Can you make some unit tests? 

Here is my draft of the final function:

```r
findCorrelation <- function(x, cutoff = 0.90, verbose = FALSE, names = FALSE, fast = FALSE) {
  if(names & is.null(colnames(x)))
      stop("'x' must have column names when `names = TRUE`")
  out <- if(fast) 
    findCorrelation_fast(x = x, cutoff = cutoff, verbose = verbose) else 
      findCorrelation_orig(x = x, cutoff = cutoff, verbose = verbose)
  out
  if(names) out <- colnames(x)[out]
  out
}


findCorrelation_orig <- function(x, cutoff = 0.90, verbose = FALSE) {
  varnum <- dim(x)[1]
  
  if (!isTRUE(all.equal(x, t(x)))) stop("correlation matrix is not symmetric")
  if (varnum == 1) stop("only one variable given")
  
  x <- abs(x)
  
  # re-ordered columns based on max absolute correlation
  originalOrder <- 1:varnum
  
  averageCorr <- function(x) mean(x, na.rm = TRUE)
  tmp <- x
  diag(tmp) <- NA
  
  maxAbsCorOrder <- order(apply(tmp, 2, averageCorr), decreasing = TRUE)
  x <- x[maxAbsCorOrder, maxAbsCorOrder]
  newOrder <- originalOrder[maxAbsCorOrder]
  
  deletecol <- rep(F, varnum)
  
  for (i in 1:(varnum - 1)) {
    if (deletecol[i]) next
    for (j in (i + 1):varnum) {
      if (!deletecol[i] & !deletecol[j]) {
        if(verbose)
          cat("Considering row\t", newOrder[i], 
              "column\t", newOrder[j], 
              "value\t", round(x[i,j], 3), "\n")  
        if (x[i, j] > cutoff) {
          if (mean(x[i, -i]) > mean(x[-j, j])) {
            deletecol[i] <- T
            if (verbose) cat("  Flagging column\t", newOrder[i], "\n")
          }
          else {
            deletecol[j] <- T
            if (verbose) cat("  Flagging column\t", newOrder[j], "\n")
          }
        }
      }
    }
  }
  newOrder[which(deletecol)]
}


findCorrelation_fast <- function(x, cutoff = .90, verbose = FALSE){
  averageCorr <- colMeans(abs(x))
  averageCorr <- as.numeric(as.factor(averageCorr))
  x[lower.tri(x, diag = TRUE)] <- NA
  combsAboveCutoff <- which(abs(x) > cutoff)
  
  colsToCheck <- ceiling(combsAboveCutoff / nrow(x))
  rowsToCheck <- combsAboveCutoff %% nrow(x)
  
  colsToDiscard <- averageCorr[colsToCheck] > averageCorr[rowsToCheck]
  rowsToDiscard <- !colsToDiscard
  
  if(verbose){
    colsFlagged <- pmin(ifelse(colsToDiscard, colsToCheck, NA),
                        ifelse(rowsToDiscard, rowsToCheck, NA), na.rm = TRUE)
    values <- round(x[combsAboveCutoff], 3)
    cat('\n',paste('Combination row', rowsToCheck, 'and column', colsToCheck,
                   'is above the cut-off, value =', values,
                   '\n \t Flagging column', colsFlagged, '\n'
    ))
  }
  
  deletecol <- c(colsToCheck[colsToDiscard], rowsToCheck[rowsToDiscard])
  deletecol <- unique(deletecol)
  deletecol
}
```

EdwinTh:
Sure, will do. Little busy now, but will look at it soon.

EdwinTh:
I was working on some unit tests for the code you provided when I found something odd. Consider the following 5 x 5 correlation matrix:

```R
> (corrMatrix <- matrix(c(1,.9,0,.4,.7,.9,1,.8,.2,0,0,.8,1,.5,.5,.4,.2,.5,1,0,.7,0,.5,0,1), 5, 5))
     [,1] [,2] [,3] [,4] [,5]
[1,]  1.0  0.9  0.0  0.4  0.7
[2,]  0.9  1.0  0.8  0.2  0.0
[3,]  0.0  0.8  1.0  0.5  0.5
[4,]  0.4  0.2  0.5  1.0  0.0
[5,]  0.7  0.0  0.5  0.0  1.0
```

If we would set the cut off at .75 than the first column to be deleted in both the original and the fast would be column 1 (since 0.9 is the highest correlation, and the mean correlation of 1 is larger than the mean correlation of 2). The difference should be in the second deletion. The fast version should delete the second column, since the average correlation before removal of column 1 is higher for column 2 than column 3. The original version should delete column 3, because after removal of column 1 it has a higher average correlation than column 2.

However this is what happens:
```R
> findCorrelation(corrMatrix3, cutoff = .75, fast = TRUE)
[1] 1 2
> findCorrelation(corrMatrix3, cutoff = .75)
[1] 1 2
```

I looked into the code of the original and indeed the average of the correlations is not recomputed after removal in a previous iteration. The reason that in your example above difference columns are removed, is because the implementations treat ties differently (note that in your example the correlations for columns 1 and 4 are exactly equal). Please check if my observation is right that the original function and the fast function do yield similar outcomes, besides the ties treatment. If that is the case then we could always use the fast option. However I would suggest that the original function is to be adapted such that correlations are recalculated. This would yield superior results to the fast implementation, which should only be use with very large correlation matrices.

topepo:
You are completely correct; the current version does not update. I found an old version (circa 2004) that I wrote but never checked in. I also added a check to see if any other elements are above the threshold for early stopping. 

The new version of the "exact" code is, on average, 26-fold slower than the current version (for 250x250 matrices) although the average time for matrices of this size was only 26s. The exact version yields subsets that are, one average, 2.4% smaller than the other versions. This difference will be more signficant for smaller matrices. The faster ("approximate") version of the code is 8-fold faster than the current version.

Here is a test case that I will add to test for the differences in the two versions:

```r
R1 <- structure(c(1, 0.86, 0.56, 0.32, 0.85, 0.86, 1, 0.01, 0.74, 0.32, 
                  0.56, 0.01, 1, 0.65, 0.91, 0.32, 0.74, 0.65, 1, 0.36,
                  0.85, 0.32, 0.91, 0.36, 1), 
                .Dim = c(5L, 5L))

exp_R1 <- c(1, 5, 4) ## with cutoff = .6
findCorrelation(R1, cutoff = .6)
```

topepo:
Just sent to CRAN. Thanks!

PR #169, opened 2015-06-15
Fixed searchpath of var_seq to caret::var_seq to make caret importable

bleutner:
Hi, 

I'm writing a package which imports caret. With the last release 6.0-47 and 0.48  all models which use var_seq in building the model list with models/parseModels.R are broken (can't find var_seq). 
I don't fully grasp why this is happening, but it seems that the import of inst/models/models.Rdata happens outside of my package's namespace and hence only works if caret has been attached before.

I created a demo package, if you want to see what's happening:

    devtools::install_github('bleutner/dummy')
    library(dummy)
    fun()
    Error in data.frame(mtry = var_seq(p = ncol(x), classification = is.factor(y),  : 
    could not find function "var_seq"

A way to solve this would be to specify the search path explicitly in the corresponding model parse files as suggested in my pull request.

I hope you can consider this and thanks very much for your awesome package
Benjamin


topepo:
How are you accessing the model code? Via the `models.RData` file? If that's the case then you should probably try using `getModelInfo` instead. I have had no issues doing that. 

I want to avoid making these changes. That function is exported (see example below) and the code is already messy enough. 

```r
library(caret)

set.seed(1)

dat <- twoClassSim(200)

modelInfo <- list(label = "Random Forest",
                  library = "randomForest",
                  loop = NULL,
                  type = c("Classification", "Regression"),
                  parameters = data.frame(parameter = "mtry",
                                          class = "numeric",
                                          label = "#Randomly Selected Predictors"),
                  grid = function(x, y, len = NULL) {
                    data.frame(mtry = var_seq(p = ncol(x), 
                                              classification = is.factor(y), 
                                              len = len))
                  },
                  fit = function(x, y, wts, param, lev, last, classProbs, ...) 
                    randomForest(x, y, mtry = param$mtry, ...),
                  predict = function(modelFit, newdata, submodels = NULL) 
                    predict(modelFit, newdata),
                  prob = function(modelFit, newdata, submodels = NULL)
                    predict(modelFit, newdata, type = "prob"),
                  levels = function(x) x$classes,
                  tags = c("Random Forest", "Ensemble Model", "Bagging", "Implicit Feature Selection"),
                  sort = function(x) x[order(x[,1]),])

set.seed(2)
rf_fit <- train(Class ~ ., data = dat, 
                method = modelInfo)
```

bleutner:
I'm simply importing caret and call train() within my functions. 

Your example works, because you have attached caret to the global search path with library(caret). When importing caret this is not the case. Also using getModelInfo() does not do the trick (neither  does loading and accessing the Rdata file manually).

I've updated the dummy package (and removed the wrong argument name for method, sorry). 
The error remains the same:

    devtools::install_github("bleutner/dummy")
    library(dummy)
    # Contents of fun()
    fun
       function () 
       {
       mod <- getModelInfo("rf")[["rf"]]
       train(Sepal.Length ~ ., data = iris, method = mod, tuneLength = 2)
       }
      <environment: namespace:dummy>
    # call function 
    fun()
    Error in data.frame(mtry = var_seq(p = ncol(x), classification = is.factor(y),  : 
    could not find function "var_seq"


As far as I can see my only other option is to move caret to Depends and force it to be attached on loading my package, which I was told, should best be avoided. Or ship my own copy of var_seq() of course, but that's an ugly workaround in my opinion,
 So I really would argue for fixing this in caret.

topepo:
Maybe it is the difference in our base install versions but it worked fine for me with the current CRAN caret version:

```
> library(caret)
Loading required package: lattice
Loading required package: ggplot2
> library(dummy)
> fun()
Loading required package: randomForest
randomForest 4.6-10
Type rfNews() to see new features/changes/bug fixes.
Random Forest 

150 samples
  4 predictor

No pre-processing
Resampling: Bootstrapped (25 reps) 

Summary of sample sizes: 150, 150, 150, 150, 150, 150, ... 

Resampling results across tuning parameters:

  mtry  RMSE       Rsquared   RMSE SD     Rsquared SD
  2     0.3686462  0.8024093  0.03170340  0.03502239 
  5     0.3574024  0.8137835  0.03246673  0.03443396 

RMSE was used to select the optimal model using  the smallest value.
The final value used for the model was mtry = 5. 
> sessionInfo()
R version 3.1.3 (2015-03-09)
Platform: x86_64-apple-darwin10.8.0 (64-bit)
Running under: OS X 10.9.5 (Mavericks)

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] randomForest_4.6-10 dummy_0.0.0.9000    caret_6.0-47        ggplot2_1.0.1       lattice_0.20-30    

loaded via a namespace (and not attached):
 [1] BradleyTerry2_1.0-6 brglm_0.5-9         car_2.0-25          codetools_0.2-10    colorspace_1.2-6    compiler_3.1.3     
 [7] digest_0.6.8        foreach_1.4.2       grid_3.1.3          gtable_0.1.2        gtools_3.4.2        iterators_1.0.7    
[13] lme4_1.1-7          MASS_7.3-40         Matrix_1.2-0        mgcv_1.8-6          minqa_1.2.4         munsell_0.4.2      
[19] nlme_3.1-120        nloptr_1.0.4        nnet_7.3-9          parallel_3.1.3      pbkrtest_0.4-2      plyr_1.8.1         
[25] proto_0.3-10        quantreg_5.11       Rcpp_0.11.6         reshape2_1.4.1      scales_0.2.4        SparseM_1.6        
[31] splines_3.1.3       stringr_0.6.2       tools_3.1.3  
```

zachmayer:
Hi Max, 

I think the problem is that the package should function as an import, without needing to load it as a separate library call.  E.g `library(dummy); fun()` vs `library(dummy); library(caret); fun()`.

Hadley et. al. are recommending new package authors import dependencies, rather than depending on them.  This is a problem for caret because we explicitly put the models OUTSIDE the caret namespace to keep CRAN happy and keep our package dependencies smaller.

Here's a simpler example of the problem (run in a fresh R session):
```{R}
caret::train(Sepal.Length ~ ., data = iris, method = 'rf', tuneLength = 2)
```
vs
```{R}
randomForest::randomForest(Sepal.Length ~ ., data = iris)
```

The random forest function runs fine, even though the package isn't loaded.

I agree that `caret::var_seq` is ugly, but I think the problem is worth fixing and I can't think of a better solution.

-Zach

bleutner:
Yes Zach, you're spot-on (even about the Hadley stuff). That's exactly the issue.

topepo:
okay, I'm convinced. 

Thanks,

Max

PR #178, opened 2015-07-06
added lfda model for initial pull request

terrytangyuan:
Hi Max, 

Could you please take a look at this pull request for addition of Local Fisher Discriminant Analysis Model (LFDA)? I haven't done any testing with the caret package yet. I am not familiar with the structure of caret so feel free to put the code into wherever appropriate for you. 

Thanks,
Yuan


topepo:
So looking at this for a while, I think it would be better to make a class out of this with a default and predict method. Bundling it inside of a model wrapper is likely to lead to some issues. 

There are templates in this (and other) packages that can help do this. For example, [`bagFDA`](https://github.com/topepo/caret/blob/master/pkg/caret/R/bagFDA.R) or [`avNNet`](https://github.com/topepo/caret/blob/master/pkg/caret/R/avNNet.R) show examples of `formula`, `print`, and `predict` methods using simple S3 code. 

Once you have that together, you could add it to this package (as long as you'll maintain the code) or start your own for the model. I'll be happy to help either way.

terrytangyuan:
Thanks! I have just tried to put them together. Could you check to see what else needs to be changed?

topepo:
It looks good. I have a few comments and questions:

* is the input `y` supposed to be a binary indicator matrix? `t(as.matrix(y))` would suggest that it is but the example above it has `y <- trainData[,1]` which suggests that maybe it isn't. 
* for the predictor data (in `x` and `newdata`), you should check to make sure that it is a numeric matrix. I usually do something like `if(is.data.frame(x)) x <- as.matrix(x)` then check to make sure that it is numeric. 
* I would move the functions defined inside of `lfda.default` outside of it and rename `lfda` to something like `lfda_calc`. 
* a default value for `r` would be good and you'll want to check for valid values of `r` and `knn` inside the function. You might need to do some sort of check that the linear algebra works in the local space (i.e. do you require that `ncol(x) > knn` for things to work?)
* someone will run a model with a single predictor, so you should defensively convert statements like `x[, y == i]` to `x[, y == i, drop = FALSE]`
* I think that the `predict` method is returning the discriminant value. The `type` option usually has values for different types of output. You might want to add some conditional statements for `type = "raw"`, `type = "class"`, and `type = "prob"`. The first could be the discriminant values, the second should be a factor based on the largest value, and the last could be computed via softmax. 
* a simple `print` method is a good idea, but we can add that later. 

Thanks,

Max

terrytangyuan:
Thanks a lot for your comments. I have fixed/added code for most of your suggestions. However, I don't get what you mean in the second last comment regarding the "type" in "predict". The "predict.lfda" method simply returns a matrix. So should it just be "raw" and then I need to add conditions to deal with situations where user put other types? 

Also, could you instruct me on how to test my functions from my own forked copy of your repo? 

If there's anything else you'd like to change, please let me know as well. Thanks so much for your time. 

Regards,
Yuan

topepo:
> However, I don't get what you mean in the second last comment regarding the "type" in "predict". The "predict.lfda" method simply returns a matrix.

So right now I think that you are returning the discriminant values for new samples (right?). Most classification functions default the predictions to be the class predictions in the form of a factor variable with the same levels as the original outcome. For example:

```r
> set.seed(1)
> dat1 <- twoClassSim(200)
> dat2 <- twoClassSim(3)
> 
> mod <- plsda(x = dat1[-ncol(dat1)], y = dat1$Class, ncomp = 3)
> predict(plsdamod, newdata = dat2[, -ncol(dat2)], type = "class")
[1] Class1 Class2 Class1
Levels: Class1 Class2
> predict(plsdamod, newdata = dat2[, -ncol(dat2)], type = "prob")
, , 3 comps

        Class1    Class2
[1,] 0.6857121 0.3142879
[2,] 0.3713097 0.6286903
[3,] 0.5945083 0.4054917

> predict(plsdamod, newdata = dat2[, -ncol(dat2)], type = "raw")
, , 3 comps

        Class1    Class2
[1,] 0.8900741 0.1099259
[2,] 0.2366989 0.7633011
[3,] 0.6913172 0.3086828
```

> So should it just be "raw" and then I need to add conditions to deal with situations where user put other types?

Yes, that's what I would suggest. 

> Also, could you instruct me on how to test my functions from my own forked copy of your repo?

I would try a lot of different data to see what happens. `twoClassSim` helps with examples on the help pages etc. We just got a new version of `caret` on CRAN, so you can test with that. Is that what you mean?

terrytangyuan:
Hi Max, 

I have done some tests on some data set. It worked as usual but the only problem is from `"lfda" <-
  function(x, ...)
    UseMethod("lfda_calc")`.
I changed this `UseMethod("lfda_calc")` to `UseMethod("lfda")` but still didn't work. Do you know the reason? I am not familiar with this `UseMethod` in R.

I had to change `out <- do.call("lfda", modelArgs)` to `out <- do.call("lfda_calc", modelArgs)` in `lfda.default` in order to make it work. 

In short, the `lfda` function does not work.

topepo:
Previously, you had a functional called `lfda` that called another function internal to that one (which I think had the same name). My suggestion was to make a class called `lfda`. You need the general function (see below) and then one called `lfda.default` [1]. I suggested taking the internal function that does the computations and defining that externally. You don't want to give it the same name, so I suggested `lfda_calc`. 

You might want something like this:

```r
"lfda" <- function(x, ...) UseMethod("lfda")
"lfda.default" <- 
  function(x, y, r = 3, metric = c("orthonormalized","plain","weighted"),knn = 5, ...)
  {
    if(is.data.frame(x)) x <- as.matrix(x)
    
    metric <- match.arg(metric)
    modelArgs <- c(list(x,y,r,metric,knn))
    out <- do.call("lfda_calc", modelArgs)

<snip>
```

It also might help to read this reference a bit [2]

[1] and probably another called `lfda.formula` but we can do that later
[2] [Advanced R](http://adv-r.had.co.nz/). See the section on "OO-essentials"

terrytangyuan:
Thanks a lot for the detailed instructions! I've fixed that and done some testing. 

Do you think this is ready to go to your repo? We'll add print, formula, etc later.

topepo:
I'll bring in the PR, but it the next step is to write the help files. The `prompt` function can help you do that.

terrytangyuan:
Sure. So in the future, how am I going to contribute? New pull request? Or am I granted access to the repo and then I create my own branch, etc?

PR #179, opened 2015-07-08
fix bug: "The tuning parameter grid should have columns nrounds, lambda, alpha"

krz:
fixes bug:
Error in train.default(x, y, weights = w, ...) :
  The tuning parameter grid should have columns nrounds, lambda, alpha


zachmayer:
@topepo Looks good to me.

topepo:
Thanks!

Issue #22, opened 2014-06-24
in predict.train, subset on features


topepo:
Right now, if extra features are in `newdata` it can crash if there was pre-processing. After the formula handling, use `predictors(object)` to filter out extra predictors. This may be `NULL`, so do some checking first.

topepo:
Added: https://github.com/topepo/caret/blob/ede9913a465078dfaec5aeff2099676857c666e6/pkg/caret/R/predict.train.R#L156

PR #213, opened 2015-08-07
Fixed issue22-deal with exception when predictors in model does not match with predictors in newdata

terrytangyuan:
This fixed issue #22

### DIFF
--- a/pkg/caret/R/predict.train.R
+++ b/pkg/caret/R/predict.train.R
@@ -21,13 +21,6 @@ predict.train <- function(object, newdata = NULL, type = "raw", na.action = na.o
       stop("only classification models that produce probabilities are allowed")
   }
   
-  if(length(object$coefnames) > dim(newdata)[2]){
-    stop("the number of predictors in newdata is smaller than that in the model")
-  } else if(length(object$coefnames) < dim(newdata)[2]){
-    warning("only the predictors contained in the model will be used for prediction")
-    newdata <- newdata[, object$coefnames]
-  }
-  
   if(!is.null(newdata))
   {
     if (inherits(object, "train.formula"))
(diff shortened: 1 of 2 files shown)


terrytangyuan:
So `object$coefnames` only extracts the names of the coefficients. I tried `predictors(object)` but it returns `NULL`. Any way I can extract all the colnames of the original training data, including the outcome y?

topepo:
Now that I've looked at this in more detail, I think that my initial thought was a bad one. 

There are a lot of models that use less predictors than those contained in the model, but the underlying model fit still needs those in the input matrix or data frame to work. This is mainly because most code records the predictors _by column number_ and not name so eliminating predictors will screw this up. 

For example, 

```r
> library(caret)
> 
> set.seed(2)
> dat <- twoClassSim(100)
> 
> set.seed(3)
> mod1 <- train(Class ~ ., data = dat, method = "rpart")
> 
> predictors(mod1)
[1] "TwoFactor2" "TwoFactor1" "Nonlinear3" "Linear04"   "Linear05"   "Linear08"  
> length(predictors(mod1))
[1] 6
> ncol(dat) - 1
[1] 15
> 
> predict(mod1, head(dat[, predictors(mod1)]))
Error in eval(expr, envir, enclos) : object 'Linear01' not found
> 
> 
> set.seed(3)
> mod2 <- train(x = dat[, -ncol(dat)], y = dat$Class, method = "rpart")
> 
> predictors(mod2)
[1] "TwoFactor2" "TwoFactor1" "Nonlinear3" "Linear04"   "Linear05"   "Linear08"  
> length(predictors(mod2))
[1] 6
> ncol(dat) - 1
[1] 15
> 
> predict(mod2, head(dat[, predictors(mod2)]))
 Show Traceback
 
 Rerun with Debug
 Error in eval(expr, envir, enclos) : object 'Linear01' not found 
```

The good news is that there are some upcoming changes to `preProcess` that will eliminate the original problem that prompted issue #22.

PR #214, opened 2015-08-10
Added 3D plot method for lfda result

terrytangyuan:


pkg/caret/tests/testthat/test_models_lfda.R
```diff
@@ -21,4 +23,14 @@ test_that('test lfda model training and prediction', {
   expect_that(dim(tr_dat)[1], equals(dim(transformed.train)[1]))
 })
 
+test_that('test lfda model visualization function', {
+  skip_on_cran()
+  set.seed(1)
+  tr_dat <- twoClassSim(200)
+
+  lfda.model <- lfda(x=tr_dat[,-16],y=tr_dat[,16],r=3)
```
zachmayer: You'll need to load the lfda package to run this test.

pkg/caret/R/lfda.R
```diff
@@ -314,3 +314,45 @@ function(object, newdata = NULL, type = "raw", ...)
...
+      }))
+    df <- sapply(newData[,"Class"], as.character) 
+    df[is.na(df)] <- "."
+    newData[,"Class"] <- df
+  } else{
+    newData <- transformedData
+  }
+
+  ## Plot 3D Visualization of LFDA Result ##
+  rgl::text3d(newData[,2],newData[,3], newData[,4], col=Cols(transformedData$Class),
```
zachmayer: If you're going to import from rgl, you need to add rgl to imports.

pkg/caret/tests/testthat/test_models_lfda.R
terrytangyuan: There isn't a lfda package. It's lfda.R in caret.

pkg/caret/R/lfda.R
terrytangyuan: Do you mean add `import(rgl)` in NAMESPACE? BTW, I added rgl in `suggests` section in DESCRIPTION file.

pkg/caret/tests/testthat/test_models_lfda.R
zachmayer: I just realized that.  At this point you really should make an `lfda` package and submit it to CRAN.  I'd be happy to help you do that.  If you initialize a github repo, I'll send you a PR to setup testthat/travis/coveralls, etc. and walk you through your first CRAN submission.

pkg/caret/R/lfda.R
zachmayer: Yes, but a better approach would be to use the roxygen `@imports` tag.  Suggests is for packages used for testing or buildign vignettes. Since this function is part of the main package, all the functions it imports need to be imported.

pkg/caret/tests/testthat/test_models_lfda.R
zachmayer: @terrytangyuan Also, I deleted my previous comment, which in retrospect I shouldn't have done.  It read "You need to load the lfda package for this test".

pkg/caret/R/lfda.R
terrytangyuan: Got it. So I should add @imports to plot.lfda.Rd and then remove rgl from suggests section in DESCRIPTION. I also should add rgl to imports in DESCRIPTION. Am I all correct?

zachmayer: `@imports lfda`, but otherwise yes, yes, and yes.

Since this is adding another dependency to caret, which we're trying to avoid, Max and I would really prefer you make a new package.

terrytangyuan: Thanks. Got it. I just initialized lfda package. Let's start collaborating on that! It's badly organized for now, compared to caret, and I am never published an R package before. So your help would be appreciated. 

So I guess I can stick with the `suggests` in DESCRIPTION for now so users won't be required to install this heavy dependency?

terrytangyuan:
lfda is on CRAN now. I'll submit a new PR for the changes.

PR #216, opened 2015-08-11
fixed travis build file

jtilly:
I fixed the travis build file after seeing your question [here](http://stackoverflow.com/questions/31879011/r-cant-find-packages-installed-by-travis). `R CMD check --as-cran` is looking for packages (that are not explicitly stated as dependency) in `/usr/lib/R/library`. I then use the "old school" travis setup to install dependencies in `/usr/lib/R/library` so they can be found. I'm also installing a bunch of dependent packages as binaries to make the build faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,6 @@ before_install:
   - echo 'R_LIBS=/usr/lib/R/library:/usr/lib/R/site-library/' > ~/.Renviron
   - sudo chmod 2777 /usr/lib/R/library
   - ./travis-tool.sh bootstrap
-  - ./travis-tool.sh r_binary_install covr
   - ./travis-tool.sh r_binary_install car
   - ./travis-tool.sh r_binary_install reshape2
   - ./travis-tool.sh r_binary_install foreach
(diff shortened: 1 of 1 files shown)


zachmayer:
Thank you for fixing this!  2 requests:

1. Fix merge conflicts.  I was also working on the .travis.yml file, but you can just resolve the conflicts by discarding all changes on master and replacing them with your file.
2. Add the coveralls line back in, and the coveralls install_github:
```{yml}
after_success:
  - Rscript -e 'library(covr);coveralls()'
```

THANKS SO MUCH!  This has been bothering me for weeks.

jtilly:
I fixed the merge conflicts and added coveralls support.

zachmayer:
THANKS!

zachmayer:
Out of curiosity, during `./travis-tool.sh bootstrap`, is it skipping the latex bootstrapping step?  I'd like to avoid latex if possible, as we're skipping manual and vignette building.

jtilly:
Yes, the current version is skipping the latex bootstrapping step. As long as the environmental variable `BOOTSTRAP_LATEX` is not set, no latex distribution will be installed.

zachmayer:
Great.  Thank you!

PR #221, opened 2015-08-16
cleanup lfda

zachmayer:
Commented out tests.  I can’t get it to make anything other than
constant predictions.


zachmayer:
Fixed the tests

PR #235, opened 2015-08-31
Running python models

cancandan:
I've added a sample model which runs `sklearn.neighbors.KNeighborsRegressor` via rPython. The python installation should have the necessary libs, ie. sklearn and pandas, which is checked in the modified `checkInstall` function. Thought might be useful


zachmayer:
Can you remove the `.DS_Store` file and add `.DS_Store` to the gitignore file?  It doesn't need to be tracked.

zachmayer:
Other than that, this seems like a good idea to me!

cancandan:
I removed .DS_Store and added to .gitignore, I tried to copy the regression test for kknn and run it for pythonKnnReg but it failed with 

`Error in trainControl(method = "cv", number = 3, returnResamp = "all",  : 
  unused argument (search = "random")` 

Don't know why

Also in the file move_files.R  there is this line `setwd("/Users/kuhna03/Code/github/caret/RegressionTests")` shouldn't it be like
 `setwd("~/Code/github/caret/RegressionTests")`

zachmayer:
kknn has a bunch of extra arguments normal knn doesn't.

cancandan:
Ok its because I have the cran release installed which does not have random search

topepo:
Great idea! 

One more request: it looks like you automatically try to install missing modules. Can you:

* do something similar to the R code install part and ask if they want to install?
* used `FALSE` instead of `F` etc?
* use `return (list())` instead of `return (NULL)`?
* add yourself to the DESCRIPTION file's contributor list?

I'll take care of the random grid search part after the PR is merged.

Thanks,

Max

topepo:
One other request: can you set the seed for the model? I'm thinking you could generate one in R (e.g. `sample.int(100000, 1)`) and pass it to python.

cancandan:
In `checkInstall` rPython is checked if its installed. If installed, python libraries are tried to be imported, if this fails a message is returned asking the user to install them and we stop. This is because the installation of the python lib sklearn is not straightforward for all platforms, unless you are using a prepackaged python like anaconda. However for example currently there's a problem with rPython linking to anaconda on osx, experiencing this I manually installed the python libs sklearn and pandas myself on top of base python, then it worked. Therefore it might be better if the user is referred to its installation page of the libs [sklearn] (http://scikit-learn.org/stable/install.html) and [pandas] (http://pandas.pydata.org/pandas-docs/stable/install.html)

zachmayer:
I agree— trying to install sklearn and pandas can be a real pain.  It's not like installing R packages unfortunately... I think it's best to tell the user to install the needed libs.

topepo:
Sounds good to me. I merged and will make some additions to the `grid` module. 

Thanks!

Max

topepo:
Can you take a look at Issue #236?

Thanks,

Max

PR #237, opened 2015-09-05
Variable importance with predictor shuffling

cancandan:
I tried to implement this interesting approach to variable importance calculation described [here](http://amunategui.github.io/variable-importance-shuffler/)
Thought may be useful.


cancandan:
Travis says:

* checking whether package ‘caret’ can be installed ... ERROR
Installation failed.
See ‘/home/travis/build/topepo/caret/pkg/caret/caret.Rcheck/00install.out’ for details.
* DONE
Status: 1 ERROR
See
  ‘/home/travis/build/topepo/caret/pkg/caret/caret.Rcheck/00check.log’
for details.

How do we check the logs?

zachmayer:
Twiddle open the "after failure" section and it should show you the log dumps.




It looks like the manual page you wrote has a mal-formatted title, but I'm on my phone and can't tell 100%.




Try running devtools::check() locally and you might get more informative errors.



—

cancandan:
Thanks for the devtools::check() tip. When I run it on master caret branch I get
```
Error: processing vignette 'caret.Rnw' failed with diagnostics:
 chunk 11 (label = rdaFit) 
Error in checkInstall(models$library) : 
Execution halted
Error: Command failed (1)
In addition: Warning message:
NAMESPACE not generated by roxygen2. Skipped.
```

pkg/caret/man/shuffleVarImp.Rd
```diff
@@ -0,0 +1,94 @@
...
+                     paste("bogus", 1:5, sep = ""))
+
+normalization <- preProcess(sim$x)
+x <- predict(normalization, sim$x)
+x <- as.data.frame(x)
+
+train_x=x[1:80,];train_y=sim$y[1:80]
+test_x=x[81:100,];test_y=sim$y[81:100]
+
+model=train(x=train_x,y=train_y,method="glmnet")
```
zachmayer: Change this to `method='glm'`.

zachmayer:
Ahah!  The problem is you're using glmnet in your example, which isn't a dependency of caret, so it's not installed on travis.  Change your example to use `glm`, `randomForest`, or another method that's listed in the [DESCRIPTION](https://github.com/topepo/caret/blob/master/pkg/caret/DESCRIPTION) file or [.travis.yml](https://github.com/topepo/caret/blob/master/.travis.yml#L26) file.

topepo:
A few thoughts:

 * it would make a more invasive change but the (scientifically) best place for this to happen is during resampling. That would be closer in function to how random forest works and it factors in the model variability (which this does not) and doesn't require a separate set of data. I'm not ready to do this in the internals yet, but maybe that is something we should think about implementing. 
 * You also might think about following the random forest approach and normalize the difference in performance by the standard deviation of performance. The discrete way that you calculate change reduces the power and requires many more shuffles. 
 * think about adding an option to show the progress bar or maybe qualify it's use with `if(interactive()` etc.
 * `calculateScore` should check the control object to see if `classProbs = TRUE`. If so, generate the class probabilities. Otherwise, a lot of model fits will throw errors. 
 * Since this is called outside of `train`, it could be easily parallelized. There is code in other parts of the package that go like this:

```r
    `%op%` <- getOper(allowParallel && getDoParWorkers() > 1)
    foreach(iter = 1:shuffleTimes) %op% {}
```

or even

```r
    `%op%` <- getOper(allowParallel && getDoParWorkers() > 1)
    foreach(iter = 1:shuffleTimes) %:%
      foreach(var = 1:ncol(data))  %op% {}
```

cancandan:
I see, great comments.
A way to make this happen with a non invasive code change would have been with [AOP](https://en.wikipedia.org/wiki/Aspect-oriented_programming). Don't know if this is possible in R. But something similar has been done in the R package knitr with hooks, which is described [here] (http://yihui.name/knitr/hooks/)

PR #238, opened 2015-09-06
Moved BradleyTerry2 to Suggested packages.

kkholst:
The BradleyTerry2 package has dependency on the lme4 package which recently started to import the nloptr package. Installing the latter can be quite challenging on specific distributed computing systems without internet access. 

I therefore suggest that the BradleyTerry2 package is listed as a suggested package (it also seems no 'standard' functionality relies on the package anyway?).


topepo:
Good idea.

Thanks.

PR #245, opened 2015-09-14
call make.names on obsLevels to properly handle spaces in factors

dondelelcaro:
If the response is a factor which has levels which contain spaces, the
out object from extractProb() will have called make.names() to change
the column names to have periods instead of spaces.

This results in

Error in `[.data.frame`(out, , obsLevels, drop = FALSE) :
  undefined columns selected

when predict is called on an object.

An eventual real fix is to set check.names=FALSE in the right bit of
extractProb() which is having this effect.

### DIFF
--- a/pkg/caret/R/predict.train.R
+++ b/pkg/caret/R/predict.train.R
@@ -53,7 +53,7 @@ predict.train <- function(object, newdata = NULL, type = "raw", na.action = na.o
                        unkOnly = TRUE,
                        ...)
     obsLevels <- levels(object)
-    out <- out[, obsLevels, drop = FALSE]
+    out <- out[, make.names(obsLevels), drop = FALSE]
   } else {
     out <- extractPrediction(list(object),
                              unkX = newdata,


topepo:
I'm going to keep the approach that we have right now, which is to throw an error (it used to be a warning) when `trainControl(classProbs = TRUE)`. This should coerce people into using levels that have valid column names. 

The reasons being are:

* there would be other code that would have to change during the model creation phase. The best place to make this change would be at the start of `train.default`. However...
* some users would have errors in their code because they were expecting certain column names and get alternate versions

This is a "damned if you do, damned if you don't" issue

zachmayer:
I agree.  I'm in favor of throwing an error very early and saying "levels must be valid R names for class probs."

That said, at some point I want to re-write parts of caret to use `data.table` to store data internally, and I've noticed that `data.tables` can have non-valid column names (e.g. names with spaces in them).

Maybe that will eventually help solve this issue.

topepo:
> I'm in favor of throwing an error very early and saying "levels must be valid R names for class probs."

I does that as of a few versions ago (I had to go back and check!)

As for data tables... how compelling would it be to use them? I've been using them a lot more lately but I don't see the potential gain for them inside of `train`. Also, we would have to be very careful about what happens inside of the (potential) parallel workers with data tables passing by reference rather than value. 

I'm not against that change but before we add another dependency it would be good to quantify the gain.

dondelelcaro:
> This should coerce people into using levels that have valid column names. 

It's really only the default arguments to data.frame() or as.data.frame() which have this limitation, though. There's no intrinsic limitation of data.frame itself which causes names with spaces to be invalid. Indeed things like: 

     data.frame(`foo bar`=1:5,bleh=6:10,check.names=FALSE)$`foo bar`

work just fine. [I suspect that the check.names default is there for historical reasons... but I'm not certain.]

topepo:
I've made some changes to make the error messages more informative rather than going down this path. 

Thanks,

Max

PR #272, opened 2015-10-12
ggplot.R

bdanalytics:
1. Fixed bug in oob scatter plot
2. Added bestTune highlight to scatter plots
3. Cleaned code for both scatter & level plots
4. Added test script for ggplot

This is my first pull request for caret, so please lmk if I need to change anything.

Thanks


topepo:
It looks great! Can you:

* add it as an option (i.e. so it isn't automatic)?
* check the cases where a scale is applied to the x-axis? For example, the highlighting works with `ggplot(mdl.svmRadial)` but disappears with `ggplot(mdl.svmRadial) + scale_x_log10()`

Thanks,

Max

PR #275, opened 2015-10-12
Ggplot: Added highBestTune as optional argument

bdanalytics:
Max:

Thanks for looking into this. Per your request, I have added highBestTune (short for highlight best tune) as an argument with FALSE as the default value.

Don't know how to modify the data for the highlight layer when a scale is applied to the x-axis. For example, with ggplot(mdl.svmRadial) + scale_x_log10(), the data for the highlight layer will need to be modified inside ggplot but ggplot will not know of the external scale ? any suggestions ? Apologies if this can be done trivially.

### DIFF
--- a/pkg/caret/R/ggplot.R
+++ b/pkg/caret/R/ggplot.R
@@ -1,5 +1,5 @@
 ggplot.train <- function(data = NULL, metric = data$metric[1], plotType = "scatter", output = "layered",
-               nameInStrip = FALSE, ...) {
+               nameInStrip = FALSE, highBestTune = FALSE, ...) {
   if(!(output %in% c("data", "layered", "ggplot"))) stop("'outout' should be either 'data', 'ggplot' or 'layered'")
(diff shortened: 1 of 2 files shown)


topepo:
The issue is in how you specify the aesthetics. I'll make the change after merging the PR but you would want to use 

```r
out <- out + geom_point(data = bstRes,
                        aes_string(x = dnm[2], y = dnm[1], color = ".col"),
                        size = 4, shape = 5)
```

instead of 

```r
out <- out + geom_point(data = bstRes,
                        x = as.numeric(bstRes[, dnm[2]]),
                        y = as.numeric(bstRes[, dnm[1]]),
                        colour = ifelse(p == 1, "red", "black"),
                        size = 4, shape = 5)
```

after creating a new variable in `bstRes ` names `.col`.

topepo:
How would you like to be listed in the contributors field?

PR #287, opened 2015-10-22
Multilayer mlp gamma svm

fjeze:
Added multiple layers to the mlp and mlpweightdecay model as separate model. Added regression test to both new models. 

Added a new tuning parameter 'gamma' to svmlinear2, modified the tuning parameter in the regression test.


topepo:
Sorry for the delay on this. Can you make one change to the SVM model code? For random search, you have 

```r
   gamma = sample(1:10, replace = TRUE, size = len))
```

That parameter doesn't have to be integer so using a uniform between two relevant values. 

Thanks,

Max

fjeze:
Your right, thats a better solution I.

I have changed both, to this:

```R
cost = 2 ^((1:len) - 3)
gamma =10^((1:len) - 5)
```

If you think this is a good solution I will make a new PR

Cheers

topepo:
Sounds fine to me.

fjeze:
Made the change is this new #303  request

PR #305, opened 2015-11-02
Fix year Scrucca's article

LluisRamon:
Hi @topepo, 

I fixed the year from Scrucca's article. From 2012 to 2013 as in [JSS](https://github.com/topepo/caret/compare/master...LluisRamon:patch-1?expand=1).

Thank you for your great work!

### DIFF
--- a/pkg/caret/man/ga_functions.Rd
+++ b/pkg/caret/man/ga_functions.Rd
@@ -67,7 +67,7 @@ The return value depends on the function.
 
 }
 \references{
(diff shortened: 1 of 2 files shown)


topepo:
Thanks. I should probably change the "XXX" too.

LluisRamon:
Probably :)

FYI: There is also another "XXX" at [safs.default.Rd](https://github.com/topepo/caret/blob/master/pkg/caret/man/safs.default.Rd)

topepo:
I just fixed them to "remove XXX content"

LluisRamon:
Great! 

Thanks.

PR #308, opened 2015-11-05
Additional params for xgbTree

terrytangyuan:
Hi Max, 

Many people have requested this so I added a couple of params to cope with changes of xgboost Github version and changed installation of xgboost back to Github. Let me know if there's any change I need to make so we can get this up ASAP. 

Thanks,
Yuan


terrytangyuan:
BTW, they are all tested on my end but may still have some mistakes.

topepo:
Doesn't `gamma` depend on the loss function? Do you know if it is standardized/normalized or what its default value is? 

Also, you might want to widen the random search ranges to explore regions that grid search might not.

terrytangyuan:
Yes, it does depend on loss function and is not standardized/normalized. The default value is not specified in the doc and hidden in the C code. From my experience, it is 0 since if you don't supply gamma, it will just split until it hits the max_depth as long as there's still gain when splitting. I have widened the range in random search and supplied some more reasonable gamma values for grid search. If one's going to have a ridiculous loss function, he will need to specify the range on his own then. Let me know if you have a better idea. 

Thanks for the careful review.

topepo:
Okay. How about defaulting it to zero for the grid search via

```r
 out <- expand.grid(max_depth = seq(1, len),
                    nrounds = floor((1:len) * 50),
                    eta = c(.3, .4),
                    gamma = 0,
                    colsample_bytree = c(.6, .8),
                    min_child_weight = c(1))
```

This will let people vary the value but it won't lead to bad performance if they don't know about it. Otherwise. I will send you all the submitted issues for when people use RMSE values that go from 0.0001 to 0.00001 =]

terrytangyuan:
Thanks. Changed.

terrytangyuan:
On green. Could you merge? Let me know if there's any other issues.

topepo:
Thanks a bunch!

daniellecroy:
Can alpha be added?

PR #318, opened 2015-11-11
Added predleaf to xgbTree for predict

terrytangyuan:
Is this the correct way to add it?

### DIFF
--- a/models/files/xgbTree.R
+++ b/models/files/xgbTree.R
@@ -18,11 +18,11 @@ modelInfo <- list(label = "eXtreme Gradient Boosting",
                                          colsample_bytree = c(.6, .8),
                                          min_child_weight = c(1))
                     } else {
-                      out <- data.frame(nrounds = sample(1:1000, size = len*10, replace = TRUE),
+                      out <- data.frame(nrounds = sample(1:1000, size = 10, replace = TRUE),
                                         max_depth = sample(1:10, replace = TRUE, size = len),         
                                         eta = runif(len, min = .001, max = .6),
-                                        gamma = runif(len*10, min = 0, max = 10),
-                                        colsample_bytree = runif(len*5, min = .3, max = .7),
+                                        gamma = runif(10, min = 0, max = 10),
+                                        colsample_bytree = runif(5, min = .3, max = .7),
                                         min_child_weight = sample(0:20, size = len*5, replace = TRUE))
                       out$nrounds <- floor(out$nrounds)
                       out <- out[!duplicated(out),]
--- a/models/files/xgbTree.R
+++ b/models/files/xgbTree.R
@@ -88,9 +88,9 @@ modelInfo <- list(label = "eXtreme Gradient Boosting",
                     
                     
                   },
-                  predict = function(modelFit, newdata, submodels = NULL) {
+                  predict = function(modelFit, newdata, submodels = NULL, predleaf = FALSE) {
                     newdata <- xgb.DMatrix(as.matrix(newdata))
-                    out <- predict(modelFit, newdata)
+                    out <- predict(modelFit, newdata, predleaf = predleaf)
                     if(modelFit$problemType == "Classification") {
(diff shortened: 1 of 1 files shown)


topepo:
What was the thought behind adding `predleaf` as an argument? I don't know of any way to pass that through from `train`? 

For the random grid search, you can get rid of the `len` multipliers on `colsample_bytree`, `nrounds`, and `gamma. 

Also, I did some testing and I don't get the same answers as the code that doesn't use the sub-model trick. Here is my testing code:

```r
library(caret)
modelInfo2 <- modelInfo
modelInfo2$loop <- NULL

###################################################################

small <- expand.grid(max_depth = c(1, 10),
                     nrounds = c(10, 100, 500),
                     eta = .3,
                     gamma = 0,
                     colsample_bytree = .6,
                     min_child_weight = 1)

###################################################################

set.seed(46)
dat <- twoClassSim(200)

seeds <- vector(mode = "list", length = 26)
seeds <- lapply(seeds, function(x) 1:40)

set.seed(1)
mod1 <- train(Class ~ ., data = dat,
              method = modelInfo,
              tuneGrid = small,
              trControl = trainControl(seeds = seeds, 
                                       savePredictions = TRUE,
                                       classProbs = TRUE))

set.seed(1)
mod2 <- train(Class ~ ., data = dat,
              method = modelInfo2,
              tuneGrid = small,
              trControl = trainControl(seeds = seeds, 
                                       savePredictions = TRUE,
                                       classProbs = TRUE))

all.equal(mod1$results$Accuracy, mod2$results$Accuracy)
summary(mod1$results$Accuracy-mod2$results$Accuracy)
mod2$times$everything[3]/mod1$times$everything[3]
```
I get:

```r
> all.equal(mod1$results$Accuracy, mod2$results$Accuracy)
[1] "Mean relative difference: 0.00694488"
> summary(mod1$results$Accuracy-mod2$results$Accuracy)
     Min.   1st Qu.    Median      Mean   3rd Qu.      Max. 
-0.005201 -0.001801  0.000000  0.001264  0.003199  0.010920 
```

They aren't large differences but they shouldn't be there.  

It looks like the results are the same for some tuning parameters, specifically when the number of rounds are small. If I re-run `mod1` with the same seed, I get the same results so that isn't the issue. 

I think the problem is in the `xgboost` package itself. For example, I would expect to get the same results doing this:

```r
> data(agaricus.train, package='xgboost')
> data(agaricus.test, package='xgboost')
> train <- agaricus.train
> test <- agaricus.test
> 
> set.seed(1)
> bst1 <- xgboost(data = train$data, label = train$label, max.depth = 2,
+                eta = 1, nthread = 1, nround = 100, objective = "binary:logistic")
[0]	train-error:0.046522
[1]	train-error:0.022263
[2]	train-error:0.007063
<snip>
> pred_1_50 <- predict(bst1, test$data, ntreelimit = 50)
> 
> set.seed(1)
> bst2 <- xgboost(data = train$data, label = train$label, max.depth = 2,
+                 eta = 1, nthread = 1, nround = 50, objective = "binary:logistic")
[0]	train-error:0.046522
[1]	train-error:0.022263
[2]	train-error:0.007063
<snip>
> pred_2_50 <- predict(bst1, test$data)
> 
> all.equal(pred_2_50, pred_1_50)
[1] "Mean relative difference: 0.0001912147"
```

I'll submit a ticket to xgboost on this.

terrytangyuan:
That's weird but fortunately it's not huge difference. I am not sure how the `seeds` works inside `trainControl` though. Specifying `predleaf` will return the predicted indices of the leaves, which would be helpful for some people/problems, e.g. use it as additional features, etc. I think passing a default `predleaf = F` would not affect the normal usage of xgboost. Let me know what you think.

terrytangyuan:
Could you re-run the Travis build? I did what you suggested but have no idea why it failed. 

Also, this might be related to your question? https://github.com/dmlc/xgboost/issues/310

terrytangyuan:
Any progress on this yet? @topepo

PR #323, opened 2015-11-16
clean commit, added multiple layers to the mlp models and gamma to sv…

fjeze:
Completely clean fork and commits


topepo:
Thanks!

Issue #349, opened 2016-01-10
Invalid seeds for boot632 train control method


asardaes:
# Simple example:

```r
data(iris)
TrainData <- iris[,1:4]
TrainClasses <- iris[,5]

knnFit1 <- train(TrainData, TrainClasses,
                 method = "knn",
                 preProcess = c("center", "scale"),
                 tuneLength = 10,
                 trControl = trainControl(method = "boot632"))
```

## Results in:
```r
Error in { : task 2 failed - "supplied seed is not a valid integer"
```
-------
# Cause:

Within the `nominalTrainWorkflow` function the nested `foreach` loops have arguments `iter = seq(along = resampleIndex)` and `parm = 1:nrow(info$loop)`. However, when `ctrl$method` is `"boot632"` the `resampleIndex` variable changes and gets the `AllData` argument appended at the beginning. This causes the `foreach` loops to do one extra iteration. 

The last element of the `seeds` list is a single numeric value, but if `length(parm) > 1` then the following code:
```r
if (!(length(ctrl$seeds) == 1 && is.na(ctrl$seeds))) 
            set.seed(ctrl$seeds[[iter]][parm])
```
... attempts to access a non-existing value, therefore doing `set.seed(NA)`, which gives the error mentioned above.

--------
# Solution:
I'm not sure about this but I think that the `seeds` list must *also* get `M` values in the last element when `ctrl$method == "boot632"`, where `M` is the number of models as specified in the documentation.

topepo:
You hit the nail on the head. It does an extra model fit on the entire data set for that method. I'll fix that and update this thread. 

Thanks,

Max

asardaes:
I checked this and it seems to work if the `seeds` parameter is not provided. Since I was working around the problem, I noticed now this doesn't work:

```r
data(iris)
TrainData <- iris[,1:4]
TrainClasses <- iris[,5]

num_resamples <- 5
num_models <- 5

# specify seeds
set.seed(329)
seeds <- sample(1000, num_resamples*num_models + 1)
seeds <- lapply(seq(1, by = num_models, to = length(seeds)), 
                function(i_seed) seeds[i_seed:(i_seed+num_models)])
seeds[[length(seeds)]] <- seeds[[length(seeds)]][1]
```

## Working

```r
# this one expects a list of length 'num_resamples' (call it L)
knnFit1 <- train(TrainData, TrainClasses,
                 method = "knn",
                 preProcess = c("center", "scale"),
[...]

PR #353, opened 2016-01-15
Better seeds (related to #349)

asardaes:
I added a couple of extra safety checks:
* Making sure there is at least one number in the last element of the `seeds` list
* Making sure no seed is `NA`

I also changed the strategy to create the seed list so that it calls `sample.int` only once. With the previous strategy there was a chance, however small, that a seed could be repeated. Unfortunately this means that any user who was using the default seed creation method would need to re-run the code to obtain the new reproducible result.

Let me know if you agree with the changes.


topepo:
Excellent. Thanks.

PR #357, opened 2016-01-19
OOB training is now reproducible if done in parallel

asardaes:
Pretty straightforward. OOB models also need setting their seeds, otherwise they're not reproducible if done in parallel.


topepo:
Thanks!

PR #368, opened 2016-02-08
Fix #365 nearZeroVar + foreach behaviour.

### LINKED ISSUE #365: nearZeroVar has problematic behaviour with foreach
Using `nearZeroVar` with `foreach` produces either errors or incorrect results as far as I can tell. The following is a simple exploration and reproducible example of the bug(s).

Assuming a simple `foreach` setup like the following:

    library(caret)
    library(snow)
    library(doParallel)
    library(foreach)
    cl = makeCluster(detectCores(), type = "SOCK")

Trying the following:

    > nearZeroVar(iris)
    integer(0)
    > nearZeroVar(iris, foreach = T)
    Error in { : task 5 failed - "dim(X) must have a positive length"

And a backtrace of the same problem:

    > options(error = recover)
    > nearZeroVar(iris, foreach = T)
    Error in { : task 5 failed - "dim(X) must have a positive length"
    
    Enter a frame number, or 0 to exit
    
    1: nearZeroVar(iris, foreach = T)
    2: foreach(name = colnames(x), .combine = c) %op% {
        nzv(x[[name]], freqCut = freqCut, uniqueCut = uniqueCut, saveMetrics = FALSE)
    }
    3: e$fun(obj, substitute(ex), parent.frame(), e$data)
    
    Selection:

This points toward a problem caused by processing individual columns rather than the whole dataframe at once. The reason seems to be the use of the `is.vector` function by `nvz` to determine whether or not its argument is a vector or a matrix. Unfortunately, `is.vector` doesn't return sane results with certain atomic types which should otherwise be vectors. For example:

    > is.vector(factor(c(1,2,3)))
    [1] FALSE

Factors, timestamps, etc, all return `FALSE` from `is.vector`. A more robust method of checking for dimensionality is using `dim` and checking for `NULL`. You can see for yourself that this is causing the problem by trying to check the variance of the `Species` column in `iris`:

    > nzv(iris$Species)
    Error in apply(x, 2, function(data) { :

You can also cast the factor into a character vector to watch it work:

    > nearZeroVar(as.data.frame(sapply(iris, as.vector), stringsAsFactors = F), foreach = T)
    integer(0)

Unfortunately this doesn't work very well, and even more dangerously, it reports incorrect results silently:

    > iris$y = rep(0, nrow(iris))
    > nearZeroVar(as.data.frame(sapply(iris, as.vector), stringsAsFactors = F), foreach = T, names = T)
    [1] "Sepal.Length"

As it turns out, besides not working on the majority of dataframes, the current implementation using `foreach` actually doesn't work in any instance. This is because `nvz` tries to pick the correct columns to report using `which`. The result, if the column happens to have zero variance, is always 1.

    > iris$x = rep(0, nrow(iris))
    > iris$z = rep(0, nrow(iris))
    > nearZeroVar(as.data.frame(sapply(iris, as.vector), stringsAsFactors = F))
    [1] 1 1 1

jcpetkovich:
In `nzv`, Instead of checking dimensionality using `is.vector`, check
for `NULL` result from `dim`.

Additionally, check the result of nzv for its length since `1` will be
returned if the vector has near zero variance, and an empty integer
vector will be returned otherwise.

### DIFF
--- a/pkg/caret/R/nearZeroVar.R
+++ b/pkg/caret/R/nearZeroVar.R
@@ -15,8 +15,11 @@ nearZeroVar <- function (x, freqCut = 95/5, uniqueCut = 10, saveMetrics = FALSE,
     res$column <- NULL
   } else {
     res <- foreach(name = colnames(x), .combine=c) %op% {
-      nzv(x[[name]], freqCut = freqCut, uniqueCut = uniqueCut, saveMetrics = FALSE)
+      r <- nzv(x[[name]], freqCut = freqCut, uniqueCut = uniqueCut, saveMetrics = FALSE)
+      ## needed because either integer() or 1, r is never 0
+      if (length(r) > 0 && r == 1) TRUE else FALSE
     }
+    res <- which(res)
     if(names){
       res <- colnames(x)[res]
     }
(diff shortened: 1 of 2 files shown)


topepo:
Thanks for figuring this out.

PR #398, opened 2016-04-01
Tuning grid fix

Laurae2:
**Eta parameter**

eta parameter is primarily dependent on nrounds in the grid. It prevents "overfitting" but you require more rounds if you intend to fiddle with eta (as the shrinkage is huge) - typical eta parameters for maximum optimization using xgbTree is below 0.001. eta should be put as close to 0 as possible in order to get the maximum from an optimized parameter set. It is invariant to any other parameters than nrounds.
Expanding the grid using nrounds + eta parameter is theoretically inconsistent as eta is the last parameter to look at, not one of the first parameters to look at. Good values for trying to find the optimal settings is usually 0.1 if you got a large amount of time, or 0.3 if you want to make it fast. Hence, as we are looking for a combination of values, 0.3 yields a good output. The user can then set it manually to 0.1 or lower with the best found parameters to increase the performance (user should use xgb.cv for that though, as it gives the exact potentially preferable nrounds to use).

**Colsample_bytree parameter**

Colsample_bytree also affects nrounds. The higher Colsample_bytree is, the higher nrounds you need. However, if you reach local CV optimum quickly, it should stop immediately at a specific round (using early.stop.round preferably, but it could take very long depending on the dataset size - here there is no early.stop.round). Here, it is not a good idea. Therefore, it is useful to remove eta parameter optimization and use colsample_bytree parameter optimization, as it can yield a higher optimization for performance metrics. Some data sets are converging with a very good metric when using 1.0, while a lot of some data sets are converging better with smaller values. As we remove one eta parameter, I added another colsample_bytree parameter to fill the hole between 0.7 and the maximum (default in xgboost) value.

**Parameter variability/dependence**

max_depth -> higher number leads to a lower amount of nrounds as it changes the tree structure (convergence is slower - there are very rare cases where it's slower to converge using smaller depth)
gamma -> higher number leads to a higher amount of nrounds (convergence is slower due to train performance metric = test performance metric when gamma gets higher)
colsample_bytree -> higher number leads to a higher amount of nrounds (due to a potentially lower access to useful features)
eta -> variant on nrounds, if eta is lower, then nrounds must be forcibly higher due to higher weight shrinkage (slower convergence when eta is low)
min.child.weight -> overfitting control parameter, same as gamma but it is mainly linked with tree depth (pruning trees whose terminal nodes have a specific lower hessian weight)

**More about eta and convergence**

eta = 0.3 -> default value of xgbTree, dependent on nrounds - fiddling eta with a so small amount of nrounds will not yield any interesting output as higher eta will pick up convergence faster. It should be left to the user to optimize eta/nrounds, as messing the pair together has a very large effect on any other parameters (because the optimization here does not stop early in case of over-overfitting). Messing with only nrounds is the better as eta depends on nrounds only, and nrounds on every other parameter (if eta is variable and nrounds fixed to an appropriate low value -> lower depth, full column subsample, no gamma, etc. should nearly always be picked up | if nrounds is variable and eta fixed -> it's a matter of optimizing the other parameters).

It is why it does not make sense to optimize eta along other variables, as it is the last parameter to look at. Typical xgb.train users are using xgb.cv with an unrealistic amount of nrounds and early.stop.round to find the best nrounds count. They optimize the parameters (except eta), then when they want to get the highest performance they drop eta to an abysmal value (like 0.001) to get the most of the best found parameters. It allows only to get to the best local converging optimum given the other parameters.

**Extra idea for parameter optimization: automatic optimal nround computation**

Potential extra idea could be adding an early.stop.round implementation.

It would be like this:
- compute the test data set
- use a watchlist using an eval_metric (can use logloss or any other evaluation metric if supplied as function...)
- set verbose = 0 to not flood the console (xgboost 0.47 -> xgb.train ignores verbose=0 due to a bug, but it's only for users who compile the latest xgboost version)
- set early.stop.round to an acceptable value (ex: 10), so it automatically stops once it converged to the test set

It would need a full rework of the functions though... xgb.cv is more appropriate also. It is possible to extract the performance metric list and the predictions from it (but not the model!!!). This way, it would automatically optimize for nrounds given the preset eta. Then, when the best parameters are found, run the model with the best parameters using xgb.train. However, overriding trainControl would be difficult. It could be possible if we assume to split the data set internally with xgb.cv (parameter nfold).

**About a comment in the code**

About: # This is a toss-up, but the # trees probably adds complexity faster than number of splits

If it is about model complexity (getting to the local optimum of the provided dataset), in most cases: number of splits is faster to add complexity, but it is also prone to not hit the local convergence optimum when the split count is too large. Number of trees usually must get very large (like ~10000) when eta is very low (~0.001). It is a matter of parameters, but splits usually leads to faster model complexity than rounds (the exceptional cases are when the splits are very low, like 5 or lower -> can be very volatile depending on the features).
If it is about computation complexity (time to compute): have to assess it, number of trees is linearly dependent, unlike number of splits depending on colsample_bytree and overfitting control parameters.

### DIFF
--- a/models/files/xgbTree.R
+++ b/models/files/xgbTree.R
@@ -13,16 +13,16 @@ modelInfo <- list(label = "eXtreme Gradient Boosting",
                     if(search == "grid") {
                       out <- expand.grid(max_depth = seq(1, len),
                                          nrounds = floor((1:len) * 50),
-                                         eta = c(.3, .4),
+                                         eta = c(.3),
                                          gamma = 0,
-                                         colsample_bytree = c(.6, .8),
+                                         colsample_bytree = c(.6, .8, 1.),
                                          min_child_weight = c(1))
                     } else {
-                      out <- data.frame(nrounds = sample(1:1000, size = len, replace = TRUE),
+                      out <- data.frame(nrounds = sample(1:400, size = len, replace = TRUE),
                                         max_depth = sample(1:10, replace = TRUE, size = len),         
-                                        eta = runif(len, min = .001, max = .6),
+                                        eta = c(.3),
                                         gamma = runif(len, min = 0, max = 10),
-                                        colsample_bytree = runif(len, min = .3, max = .7),
+                                        colsample_bytree = runif(len, min = .3, max = 1.),
                                         min_child_weight = sample(0:20, size = len, replace = TRUE))
                       out$nrounds <- floor(out$nrounds)
                       out <- out[!duplicated(out),]


topepo:
I don't  agree that the range of values tested by random search (esp `eta`) should be narrowed. The purpose of random search is to cast a very wide net in the parameter space. For `eta`, we could have a wide range but bias it in a specific direction (or towards a target value).

Laurae2:
> For `eta`, we could have a wide range but bias it in a specific direction (or towards a target value).

When tuning `eta`, you must put it to abysmal low values like `0.00001` in order to get the maximum performance of the model. It affects only the computing time (as you need to increase `nrounds` when decreasing `eta`) and the ability of the model to hit the best performance ceiling. The best value of `eta` is obviously the closest to 0 you can get. It only depends about how much acceptable time you can spend waiting for the computation.

In any case, given any invariant combination of parameters [X], if `eta1 < eta2`, the best general performance when `nrounds` is (automatically optimized) set is: `[X with eta1] < [X with eta2]`. You will always get a better model when you lower eta.

It goes like this for xgboost (tree) parameter tuning:
- You tune the parameters which are not `eta`/`nrounds` by using an `infinite nrounds` + `early.stop.round` + "large" `eta` in order to halt the model when the parameters used converge to a local performance minimum (how large `eta` is depends on the data set and the time you are ready to wait)
- When you found the optimized parameters for your data set, you need to tune `eta`/`nrounds`.
- To find a close appropriate `nrounds` according to `eta`, you must cross-validate with `early.stop.round` in order to halt computations when you CV performance decreases.

`eta` is not an "optimizable" parameter (we know it must be set the closest to 0 as possible), it depends only on how much computing power and how much time you have to optimize the other parameters. When you got the optimal parameters, you put `eta = <some very low value>` and it automatically pushes to the (closest, because `eta` can't be 0) limit of the model provided the other parameters.

If you are taking `eta` as a parameter to optimize, you are slowing down the computation (because `nrounds` must be increased - example: `eta = 0.001` and `nrounds = 15000` are appropriate when you got hours, if not days or weeks to train a model). You also add one (non-needed) parameter to optimize, as the optimal `eta` parameter is known from the start.

Concrete example: suppose you are trying to find the optimal parameters of `depth`, `gamma`, `eta`, and `nrounds`.

-> `eta`/`nrounds` must not be used as they are not parameters to optimize, therefore you must optimize first `depth` and `gamma`
-> `nrounds` is not to optimize as it gets optimized by itself, you must use ~infinite values of `nrounds`, such as `1000000`
-> You must set a specific `eta` value, if you are trying to compute fast (and accept a potential local error) => 0.3 is acceptable (or higher, it depends on the size of the data set), else if you have plenty of time (or if you can't accept not having a parameter error of only 1 value) => 0.1 or even lower
-> You tune `depth`/`gamma` together using `early.stop.round` in order to find the best combination of `depth`/`gamma` giving the best performance
-> Provided the best combination of parameters, you drop `eta` as much as possible depending on the computing resource/time you are given

However, if it was other overfitting control parameters such as `gamma`, I would have agreed it would not be appropriate to remove them from tuning (especially as we do not know their optimal values). Fixing `eta` value during parameter optimization is more for computation convenience than for pushing the model to the limit. You push xgboost to the limits only when you found the optimal parameters (other than `eta`/`nrounds`).

I am not sure if you have looked how xgboost works in depth, but eta shrinks the weight of trees in order to slow down how fast it converge (how fast it takes steps).

For instance (using a metaphor here), if you are trying to find the best step 7.605 (assuming under 7.605 is undefitting, and over 7.605 is overfitting):
- stepping 1 with `eta = 1`, you will end up at step 8 (you overfit train, therefore you lose CV performance) or 7 (you largely underfit train)
- stepping (approximately) 0.01 with `eta = 0.01`, you will end up at step 7.6 (you slightly underfit train, but you give a much better CV performance)

topepo:
Sorry to leave this open for so long.  To be honest, I don't agree with a lot of what you are saying about the parameters and how they should be optimized.

PR #428, opened 2016-05-16
predict.train documentation fix for type="prob"

shanebutler:
Documentation for predict.train() references type="probs" but the actual argument should be type="prob", per the rest of the documentation file.


topepo:
Thanks!

Issue #431, opened 2016-05-25
Change default arguments of ROC computation (pROC) in the RFE process (caret)


RafaOR:
I am computing a SVM-RFE model using "ROC" as the metric with the `rfe` function of the `caret` package and I have noticed that there is no way to change the default arguments of the `roc` function (`pROC` package). In my case, for example, I would like to set the direction argument to `"<"` instead of `"auto"` because in some cases the resulting AUC is computed in reverse. Would it be possible to consider this enhancement?

xrobin:
This is actually an issue with worse classifiers with an AUC is close to 0.5. The AUC is systematically over-estimated with the default direction (auto) that will attempt to keep the AUC above 0.5 at all times. In general one should never use the default direction in a resampling operation.

The way I handle this in pROC with the bootstrapping operations is to compute the direction on the whole ROC curve, and keep that direction fixed during the resampling.
An alternative would be to simply fix the direction to "<". It is reasonable most of the time actually, as positive classifications are typically assigned to higher values. I am not very familiar with caret, so I am somewhat unsure if we can assume it will always be the case, but even if it is not it is probably better than the current state.

The ROC function from pROC seem to be called in 3 different places, in aaa.R and filterVarImp.R. I will try to submit a pull request with the latter approach.

topepo:
We need to be careful here. Right now, we assume that the first factor is the event of interest. That has been consistent in the documentation and teaching materials for some time. The changes should ensure that this design choice is not negated.

xrobin:
And pROC assumes that the level of interest is the second level of the factor, so we already have an inconsistency indeed, and relied on the direction="auto" argument. This is probably the biggest design mistake I did with pROC.

Ideally it would be best to have calls like this:

    levelOfInterest <- lev[1]
    levelOfNegatives <- lev[2]
    pROC::roc(obs, data[,class], direction = "<", levels = c(levelOfNegatives, levelOfInterest))

So two things are left:

* The levels are not available in rocPerCol. Can I assume that I will get them correctly if I do

    lev <- levels(cls) or could they end up being reversed?

* Do you always assume that the observations of interest will get higher predicted values, or could this be false in some cases?

topepo:
Here is how `rocPerCol` is used in `filterVarImp` (where the original outcome factor may have many levels):

```r
for (i in 1:k) {
  for (j in i:k) {
    if (i != j) {
      classIndex[[i]] <- c(classIndex[[i]], counter)
[...]

PR #432, opened 2016-05-25
Fix direction argument for pROC (#431)

xrobin:
First attempt to address issue #431


zachmayer:
You should add a testthat test

xrobin:
I guess that means adding tests for the twoClassSummary, multiClassSummary and rocPerCol functions? The last one is internal, do you have any recommendation for the tests?

xrobin:
I pushed a simple test for rocPerCol. It fails with the current version of caret, although I couldn't see it with an R CMD check. I'll add tests for the two other functions later on.

topepo:
Thanks.

PR #433, opened 2016-05-29
preProc updated to preProcess

washcycle:
I didn't see any backwards compatibility, but I believe this is the proper parameter name for the train.default function now.
-Matt


topepo:
Thanks. The option hasn't changed; R will do partial matching on the argument name. While I will never say "thx" instead of "thanks" I have almost always used `preProc` instead of `preProcess` :-/

washcycle:
I didn't know that (or more than likely have forgotten), I learned (or relearned) something. 

thx :P

PR #446, opened 2016-06-27
Added forceRefit option in update.train - Fixes #445

### LINKED ISSUE #445: Correct Way of Using update.train?
@topepo Is this the correct way to continue training from an existing model? Seems like `update` only updates the model object to some most recent compatible version without actually calling fit using the new dataset? I think I am missing something important here. 

```
previousModel <- readRDS("....xxx.rds")  # load previously saved model object
previousModel$trainingData <- trainData # change training data to new data
newModel <- update(object = previousModel)
```

terrytangyuan:


terrytangyuan:
@zachmayer

pkg/caret/R/updates.R
```diff
@@ -1,6 +1,6 @@
-update.train <- function(object, param = NULL, ...)
+update.train <- function(object, param = NULL, forceRefit = F, ...)
```
zachmayer: Let's spell it out all the way: `forceRefit = FALSE`

terrytangyuan: Good catch!

zachmayer:
It looks like if `param = NULL` and `forceRefit=TRUE`, you'll need to extract the tuning parameters used by the final model for the retraining, e.g. `param <- object$bestTune`.

You could also force this behaivor in the current version of caret with `new_model <- update(old_model, param = old_model$bestTune`

terrytangyuan:
Thanks for pointing out. But I don't see any success on my end though. After `update` is being called. It just finished instantly.

terrytangyuan:
@zachmayer ping! updated code.

pkg/caret/R/updates.R
```diff
@@ -1,5 +1,6 @@
-update.train <- function(object, param = NULL, ...)
+update.train <- function(object, param = NULL, forceRefit = FALSE, ...)
 {
+  if(forceRefit) param <- object$bestTune
```
zachmayer: I think this should be `if(forceRefit & is.null(param))` or even `if(is.null(param))`

zachmayer: Actually, let's go with `if(forceRefit & is.null(param))`

zachmayer:
Needs a test though

zachmayer:
LGTM.  needs a review from @topepo

terrytangyuan:
@topepo Could you take a look?

PR #448, opened 2016-07-02
Update lm.R

schistyakov:
Add ability to build model without intercept. Previous version always used intercept and there was no any way to build the model without it.

The documentation of `LM` says (see details):
>A formula has an implied intercept term. To remove this use either y ~ x - 1 or y ~ 0 + x. 

New parameter `intercept` is logical (TRUE/FALSE) 
TRUE - model with intercept (as the original version)
FALSE - model without intercept (was not possible to do it in caret)

### DIFF
--- a/models/files/lm.R
+++ b/models/files/lm.R
@@ -6,7 +6,7 @@ modelInfo <- list(label = "Linear Regression",
                                           class = "logical",
                                           label = "intercept"),
                   grid = function(x, y, len = NULL, search = "grid") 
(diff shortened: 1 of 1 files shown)


topepo:
I'd rather default the intercept to `TRUE` so that it doesn't break existing code. Can you update?

schistyakov:
Thanks for the good advice. I've just updated the default to TRUE

topepo:
Thanks!

PR #454, opened 2016-07-07
Add tuning params to rlm model: intercept, psi

schistyakov:
Add 2 tuning params to rlm model
`intercept`: If intercept needed. Logical. Should be one of `TRUE`, `FALSE`
`psi` - psi function for the model. Character. Should be one of `psi.huber`, `psi.hampel`, `psi.bisquare`


topepo:
Thanks

PR #475, opened 2016-08-24
PPV +  NPV simplified. Complicated expression replaced by equivalent and simpler

stefansimik:
I would suggest to replace the complicated expression for PPV and NPV (positive and negative predictive value) with simpler one and numerically equivalent.


topepo:
Those formulas are only the case when the prevalence is derived from the data and are not true generally. Plus, we let the user optionally specify the prevalence. 

Thanks though

stefansimik:
Thank you, clear ;)

Issue #468, opened 2016-08-10
RWeka classifier hyperparameters


dashaub:
Is there a reason that some of the `RWeka` classifiers only have one specified hyperparameter?  The LMT model has multiple available values for hyperparemeters for grid/random search while others do not. For example, J48 has `C = 0.25`, PART has `threshold = 0.25` and `pruned = "yes"`, and JRip has `NumOpt = 1:len`.

```
library(caret)
data(iris)
library(RWeka)

# No other hyperparameter values, no random search
getModelInfo("J48")

# No other hyperparameter values, no random search
getModelInfo("PART")

# This needs sampling for random search
getModelInfo("JRip")
# Also it appears NumOpt it now named O in RWeka
RWeka::WOW("JRip")

# Other values of these hyperparameters run when called directly through RWeka

# J48 called through RWeka
j48c1 <- J48(Species ~ ., data = iris, control = Weka_control(C = 0.2))
j48c2 <- J48(Species ~ ., data = iris, control = Weka_control(C = 0.25))

# JRip called through RWeka
jripc1 <- JRip(Species ~ ., data = iris, control = Weka_control(O = 3))
jripc2 <- JRip(Species ~ ., data = iris, control = Weka_control(O = 100))

# PART called through RWeka
partc1 <- PART(Species ~ ., data = iris, control = Weka_control(threshold = 0.25, pruned = "yes"))
partc2 <- PART(Species ~ ., data = iris, control = Weka_control(threshold = 0.2, pruned = "no"))
partc3 <- PART(Species ~ ., data = iris, control = Weka_control(threshold = 0.25, pruned = "yes"))
partc4 <- PART(Species ~ ., data = iris, control = Weka_control(threshold = 0.2, pruned = "no"))
```

I can do some testing on what are decent values for these hyperparameters and put together a PR if necessary.

topepo:
I don't think that the confidence values really does anything (useful). Feel free to make a PR though; I'd be happy to have more parameters for these models.

dashaub:
It looks like it might do something on the `spam` dataset:
```
library(kernlab)
data(spam)
set.seed(34)
possiblecValue <- seq(from = 0.001, to = 0.5, length.out = 100000)
numModels <- 50
cValue <- sample(possiblecValue, numModels)
pctCorrect <- MAE <- Kappa <- rep(0, numModels)
for(i in 1:numModels){
  print(i)
[...]

PR #477, opened 2016-08-26
Add RWeka Tuning Parameters

dashaub:
Adding the tuning parameters as discussed in issue #468 . Some notes:
- See issue #468 for the model performance results of before and after.
- Tuning grids for many models did not implement a random search previously. Random search is now implemented, and the grid search is also improved to test more than a single model.
- Tuning grids for PART and J48 only tested `C = 0.25` previously. We now sample in `C` in [0, 0.5].
- Sampling for many tuning parameters is set to a maximum `upperBound <- min(50, nrow(data) / 2) ` to prevent models failing to train; even for very large `nrow(data)`, performance when the tuning parameters were set to >50 was uniformly poor. Furthermore, testing showed that although this was the upper possible values for which the model would train, optimal models usually had much lower values, so sampling is done on the log scale instead of uniform `1:upperBound` to increase the probability of each model fit being in a good region for that tuning parameter.
- I ran `parseModels.R` for testing, but the updated `models.RData` object isn't included in this PR, so you'll have to run this on your end.
- The grid search construction for JRip is a little funny since we need to sample integer values on three tuning parameters. If you can think of a cleaner way to implement sampling on an integer grid here, go for it. Ultimately I decided to sample effectively on two of them and vary `NumFolds = 2:4` since that is usually a good range for `NumFolds`.
- "Collisions" for some of the random tuning grids generated on whole numbers--not a problem when we're sampling a continuous instead of a discrete numbers--are resolved by generating additional models, selecting `unique()` models, and limiting to the desired `len` with subsetting.

### DIFF
--- a/models/files/J48.R
+++ b/models/files/J48.R
@@ -8,8 +8,8 @@ modelInfo <- list(label = "C4.5-like Trees",
                   grid = function(x, y, len = NULL, search = "grid"){
                     upperBound <- min(max(1, floor(nrow(x) / 2)), 50)
                     if(search == "grid"){
-                      out <- expand.grid(C = seq(0.01, 0.5, length.out = len / 2),
-                                         M = 1:upperBound)
+                      out <- expand.grid(C = seq(0.01, 0.5, length.out = len),
+                                         M = 1:min(upperBound, len))
                       if(len == 1){
                         out <- data.frame(C = 0.25, M = 2)
                       }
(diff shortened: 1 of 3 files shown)


topepo:
One request... when using `J48`, can you add some logic to make `C = .25` when `tuneLength = 1`? If would be a good idea to enable people to use the default that Quinlan suggests easily.

dashaub:
Good idea. I also updated the default values for `PART` and `JRip` to their WEKA defaults for when `tuneLength = 1` and fixed the grid sorting to sort on the added tuning parameters.

dashaub:
Added some fixes and cleanup to the grid search tunelength.

PR #497, opened 2016-09-29
Fix undefined cross-reference

benjamfield:
Hello,

I think this should fix an undefined cross-reference in the compiled PDF ("To tune a model using Algorithm ?? ...")

Thanks


topepo:
That is missing. Should we be keeping the `{train_algo}` part?

benjamfield:
Yes, you're correct! Sorry about that.

topepo:
Thanks!

PR #484, opened 2016-09-10
Converting package to use Roxygen

JackStat:
I about went crazy on this one until I found R2roxygen. Hopefully this is useful

PR #506, opened 2016-10-12
Fixes for check problems relating to usage documentation

JackStat:
This will take some time do to the tedious nature but I wanted to get the initial commits up so you can see what cause the problem. Still needs a number of fixes before it will be ready.

See #484 comments


topepo:
Thanks

JackStat:
Just FYI I wasn't quite finished with that. Were you planning on taking it from here or should I continue?

topepo:
I'm a bit buried so, if you could, that would be great. Otherwise let me
know and I'll go after the other part sin a week or so.

PR #510, opened 2016-10-21
Fixing roxygen issues in RMD Check

JackStat:
Please don't merge yet


JackStat:
For the SBF type functions (caretSBF, lmSBF, etc.) they are a list and not a function but they each contain a function. This is causing some conflicts in the documentation because the params are in the function and not the list.  Would it break if we made them functions that returned lists?

topepo:
> Would it break if we made them functions that returned lists?

Yeah, that would break things. 

They are static in the sense that their "syntax" isn't going to change. Can we create a separate non-oxogen Rd file that contains these?

JackStat:
Found what I needed to do. I just needed to point the anovaScores function to that set of documentation like it used to be

topepo:
Let me know when you want me to merge. I've tracked down a few missing exports for `resamples.R` that crash the vignettes.

JackStat:
Not sure how to fix these...

    Undocumented arguments in documentation object 'caret-internal'
      ‘x’ ‘y’ ‘wts’ ‘method’ ‘tuneValue’ ‘obsLevels’ ‘pp’ ‘last’ ‘sampling’
      ‘classProbs’ ‘...’ ‘prefix’ ‘items’ ‘data’ ‘smoother’ ‘cut’ ‘pred’
      ‘obs’ ‘modelFit’ ‘newdata’ ‘preProc’ ‘param’ ‘ind’ ‘testX’ ‘testY’
      ‘sbfControl’ ‘object’ ‘top’ ‘names’ ‘iter’ ‘start’ ‘exclude’ ‘fixed’
      ‘seq’


    Documented arguments not in \usage in documentation object 'twoClassSummary':
      ‘formula’ ‘na.rm’ ‘x’
    ... 27 lines ...

There are also a number of these types of warnings

    S3 methods shown with full name in documentation object 'summary.bagEarth':
      ‘summary.bagEarth’

topepo:
Awesome. Thanks for all of the work. I'll merge

PR #512, opened 2016-10-31
Deal with missing data in xgbTree model using convention that missing = NA.

brockf:
This allows users to tune xgbTree models using caret for data that contain missing values. Previously, these models failed to tune because xgb.DMatrix() throws an error if you try to prep a matrix with missing values without having indicated `missing = NA`.


topepo:
Excellent. Thanks!

jianqin123:
how can I use this to solve the same problem?source the changes file doesn't work!

jianqin123:
How can I use this to solve the same problem?source the changes file doesn't work

jianqin123:
How can I use this to solve the same problem?source the changed file doesn't work

topepo:
You would have to produce a reproducible example and the results of `sessionInfo()` to get more information.

PR #520, opened 2016-11-07
Recovering the missing S3 methods

JackStat:
I went through the namespace before roxygen and found all of the s3 methods that were not in the current namespace.  I then went through and exported them all correctly


topepo:
Thanks again. It looks like you inherited the ones that I did earlier today.

JackStat:
Sorry for the mess.  I should have been much more careful

topepo:
Well, I'm the one that sent it to CRAN :-/

I think that I'm ready to resubmit. Is there anything else that you think that we should do before that?

JackStat:
I can't think of anything. Next steps for me will be a bunch of testing code so we can avoid this sort of stuff

PR #532, opened 2016-11-23
Fix missing skip parameter and fix timeslice documentation

vspinu:
a3a9963 didn't add `skip` to the `train`. Fix that. Also document `timeslice` method in `trainControl` and add a bunch of extra regexps to .gitignore.


topepo:
Thanks

PR #533, opened 2016-11-26
Fix some links to online documentation

vspinu:


vspinu:
I think these are pretty much all within the package documentation. Those within the html documentation should be tackled separately. 

I also don't see where `http://topepo.github.io/caret/sampling.html` should be re-pointed.

topepo:
Thanks for doing this. The html directory isn't used anymore.

PR #535, opened 2016-11-26
Optimize createTimeSlices and add tests

vspinu:
Current implementation of createTimeSlices is rather unfortunate. It creates full slicing with increment 1, and only after subsets with `skip=n`. Even on moderate data sets of 100k and `skip=1000` this blows the memory to 10GB while the final object is about 5MB. Fix this and add tests. 

The new algorithm generates exactly same slices as before. The only difference is the naming. Now subsamples are named with the end index of the train subsample (which is also the starting - 1 index of the test subsample). These are meaningful and always distinct. Previous naming wasn't particularly meaningful.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,8 @@ RegressionTests/.RData
 
 caret.Rproj
 
+auto/
+
 release_process/talk/caret.vrb
 
 release_process/talk/caret.nav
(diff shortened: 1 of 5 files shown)


topepo:
Thanks. It looks good

PR #536, opened 2016-11-27
Fix all broken links in bookdown files

vspinu:
I have semi-automatically fixed broken links in bookdown files and added a couple of missing redirects to `bookdown/redirects/` folder. 

The only link that I couldn't fix is `notes.html`. Where should it go?

I am relying on online docs myself. Would you please update them once you pull this? Thanks.


topepo:
I'll start remaking the pages but it will take a few days; I'm changing jobs and short a server.

PR #546, opened 2016-12-05
Fixed probValues for optimism bootstrap

asardaes:
I tried to test different scenarios, including trainings with submodels (with `earth`), but let me know if I missed something.


topepo:
It's probably just (bad) luck that I tested with SVMs.

PR #547, opened 2016-12-06
Fix classification in optimism bootstrap

asardaes:
Hello again. Sorry for all the pull requests, I could've sworn I tested more cases locally and they all worked, but apparently I was half asleep at the time. 

The fix in my previous commit broke classification with optimism bootstrap when `classProbs = FALSE`. I've fixed it and I tested both classification and regression. It seems to be working on my end. Let me know if it works for you too.

### DIFF
--- a/pkg/caret/R/workflows.R
+++ b/pkg/caret/R/workflows.R
@@ -220,7 +220,7 @@ nominalTrainWorkflow <- function(x, y, wts, info, method, ppOpts, ctrl, lev, tes
           colnames(probValues) <- lev
           if(!is.null(submod))
           {
-            probValues <- rep(probValues, nrow(info$submodels[[parm]]) + 1L)
+            probValues <- rep(list(probValues), nrow(info$submodels[[parm]]) + 1L)
           }
         }
         if(testing) print(head(probValues))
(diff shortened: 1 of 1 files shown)


topepo:
There is a branch that is being used to make a `train` methods for the [`recipes`](https://github.com/topepo/recipes) package. 

There is a [github project page](https://github.com/topepo/caret/projects/1) for these activities but mentions there do not appear to notify users. 

Can you check the recipes workflow in terms of the bootstrap optimism code?

asardaes:
I could have a look, I assume [this is the entry point](https://github.com/topepo/caret/blob/recipe_integration/pkg/caret/R/recipes.R#L927)?

If I get it correctly, the goal is to change the code that calculates the `*Extra` values so that it uses the new functions, correct?

(Just to confirm, I had not received any notifications)

topepo:
Yes, that's the function to look at. 

>  the goal is to change the code that calculates the *Extra values so that it uses the new functions

I don't know that you would need to change anything. If you could take a look and see if there are any places where they interact. The data is passed in (and processed) differently using recipes but my intuition is that this resampling shouldn't be impacted. 

Thanks!

asardaes:
I see for example [here](https://github.com/topepo/caret/blob/recipe_integration/pkg/caret/R/recipes.R#L1044) the `rec_prob` function is now used, but `probValuesExtra` right below it could probably use `fill_failed_prob` (I'm guessing right now). That's why I thought some code has to be adjusted.

Also, I see that in some places there are calls to `getFromNamespace` (for example [here](https://github.com/topepo/caret/blob/recipe_integration/pkg/caret/R/recipes.R#L1082)), is that because of `foreach`? You could simply specify the necessary functions in the `.export` parameter when calling `foreach`, but this has to be done for any internal function that is not exported (e.g. `foreach(..., .export = c("outcome_conversion"))`). That way you don't have to export internal functionality just so that it is visible in the parallel loop.

topepo:
In regards to `fill_failed_prob` and the like, there are probably a lot of places to make the code cleaner (I've been trying to refactor a bunch) so please do. Also, if you see anything else in the workflows that can be modularized (or adjust the existing functions) let me know. 

 `getFromNamespace`... yeah that is mostly do to `foreach` on specific platforms. The parallel backends are really different in how they carry along the environment from the master process so we re-import a lot of stuff. Some of that is paranoia but there is/was good reason to be paranoid.  One thing on my list is to make a local copy of the results of `getFromNamespace("outcome_conversion", "caret")` inside the foreach loops to make the code cleaner and do avoid duplication. 

Never look at code that you wrote 12 years ago :-/

Issue #549, opened 2016-12-08
xgbTree with classProbs = TRUE is completely broken on current (Dec 7) github master branch


hadjipantelis:
Hello,

This is directly related to [511](https://github.com/topepo/caret/issues/511) and the submit [d992fad](https://github.com/JackStat/caret/commit/d992fadc69dc7fdab3342cb1ab350ab95105ebdf) and the related commit [ed79d6d](https://github.com/topepo/caret/commit/ed79d6d41fe8cef0682a58527a6e3f85c5d3fbc0); it summarily kills `xgboost` when setting ` classProbs = TRUE`. All submits after that are broken for method `xgbTree`.

``` R
remove.packages('caret')
devtools::install_github('topepo/caret/pkg/caret')
library(caret)
 
set.seed(42)  
trainData <- twoClassSim(500)    
testData <- twoClassSim(500)  

trCntlList <- trainControl(method = "boot", number = 6, classProbs = TRUE) 

set.seed(55)  
xgbFitError1 <- train(  Class ~ ., data = trainData,  objective = "binary:logistic", 
                   method = "xgbTree", tuneLength = 2, trControl =  trCntlList)
set.seed(55)  
xgbFitError2 <- train( Class ~ ., data = trainData, objective = "binary:logitraw",
                   method = "xgbTree", tuneLength = 2, trControl =  trCntlList) 

trCntlList <- trainControl(method = "boot", number = 6, classProbs = FALSE) 

set.seed(55)  
xgbFitOK1<- train(  Class ~ ., data = trainData,  objective = "binary:logistic", 
                   method = "xgbTree", tuneLength = 2, trControl =  trCntlList)
set.seed(55)  
xgbFitOK2 <- train( Class ~ ., data = trainData, objective = "binary:logitraw",
                   method = "xgbTree", tuneLength = 2, trControl =  trCntlList) 
```

The error occurs within the `foreach` loop of `nominalTrainWorkflow`. In particular around Line # ~203 during the call to the `probFunction`. `probFunction` errs because the argument `modelFit$param$objective` is NULL. To that respect, `modelFit` does not have a `param` field to begin with.  That is somewhat surprising but that it can be explained by the fact that `createModel` never appended a `param` field. This is expected as `createModel` is fully agnostic about the objective function used. It should have a point where it creates a `params` field to the `modelFit` returned.

[...]

PR #550, opened 2016-12-08
Fix for issue #549

hadjipantelis:
It corrects the issue #549.


hadjipantelis:
I am cancelling this because it while solving the issue in #549 I think it is a bit premature because it does not check if the parameters field is empty (ie. the user does not define the obj. function explicitly).

PR #556, opened 2016-12-21
point to caret GitHub repo

znmeb:
The "github-repo" tag in index.Rmd needs to point to the caret repo, not bookdown-demo

### DIFF
--- a/bookdown/index.Rmd
+++ b/bookdown/index.Rmd
@@ -9,7 +9,7 @@ output:
       toc:
         collapse: section
 documentclass: book
-github-repo: rstudio/bookdown-demo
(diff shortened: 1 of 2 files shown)


znmeb:
More fixes:

1. `type = "both"` doesn't work on Linux. See <https://github.com/znmeb/caret/issues/1>. I've added guards to call installers without `type = "both"` on Linux.
2. The loop to install missing packages at the end doesn't find a number of the packages because they're in BioConductor, not CRAN. So I use `biocLite` in the loop instead.

Note also that the `fresh` switch is used before it's defined. I didn't change that; I set `fresh` to TRUE before running the script.

PR #558, opened 2016-12-22
Fixes to update_pkgs.R

znmeb:
1. `type = "both"` doesn't work on Linux - see <https://github.com/znmeb/caret/issues/1>. I added some guards to the calls.
2. The loop at the end to install missing packages wasn't finding a bunch of them. They're in BioConductor, so I changed the `install.packages` in the loop to `biocLite`.

Note also that the `fresh` variable is used before it's defined. I didn't fix that; I just set it to `TRUE` before sourcing the script.

### DIFF
--- a/bookdown/index.Rmd
+++ b/bookdown/index.Rmd
@@ -9,7 +9,7 @@ output:
       toc:
         collapse: section
 documentclass: book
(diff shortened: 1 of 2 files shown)


topepo:
Thanks

PR #569, opened 2017-01-09
return full list of variables in xgb*$varImp

ilia-kats:
xgb.importance returns only features which are used at least once in the classifier. While usually not a problem, missing features in the importance data frame lead to an error in rfe when using caretFuncs with xgbTree as classifier. This patch adds the missing features to the importance output.


topepo:
Thanks!

Issue #587, opened 2017-02-01
multiClassSummary of binary classification errs (unsurpringly but quite harsly) 


hadjipantelis:
Hello,

Using `multiClassSummary` for a binary classifier is probably a bit redundant but `train` fails extremely inelegantly and uninformatively when this happens. 

```R
rm(list=ls())
devtools::load_all('/home/pantelishadjipa/git_projects/caret/pkg/caret/')
N = 1010;
M = 10;
set.seed(1)
xTrain = matrix( runif(N*M), nrow = N);
colnames(xTrain) <- 
  sapply(1:M, function(u) paste0(collapse = '', letters[sample(26, 3, replace = TRUE)] ) )
yTrain = as.factor( letters[sample(2, N, replace = TRUE)])

trCntlListMulti  <- trainControl(method = "repeatedcv", verboseIter =TRUE, 
                                 classProbs = TRUE, 
                                 summaryFunction = multiClassSummary ) 
 
enFitMulti <- train( x = xTrain, y = yTrain, trControl = trCntlListMulti,  
                     method = "glmnet", tuneLength = 4 ) #  Errs

# Error in vector(type, length) : 
# vector: cannot make a vector of mode 'NULL'.
```
It would be nice to either disallow `multiClassSummary` in the case of binary data or at least give some sort of more informative error. (As a somewhat more far reaching feature maybe `multiClassSummary` should automatically work for binary data)

All best, 
Pantelis  

```R
> sessionInfo()
R version 3.3.1 (2016-06-21)
Platform: x86_64-redhat-linux-gnu (64-bit)
Running under: CentOS Linux 7 (Core)

locale:
 [1] LC_CTYPE=en_GB.UTF-8       LC_NUMERIC=C               LC_TIME=en_GB.UTF-8        LC_COLLATE=en_GB.UTF-8    
 [5] LC_MONETARY=en_GB.UTF-8    LC_MESSAGES=en_GB.UTF-8    LC_PAPER=en_GB.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_GB.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[...]

PR #596, opened 2017-02-11
zapping issue #587

JackStat:
It seems that it was a naming issue that was pulling in a column named NA

### DIFF
--- a/pkg/caret/R/postResample.R
+++ b/pkg/caret/R/postResample.R
@@ -224,8 +224,6 @@ multiClassSummary <- function(data, lev = NULL, model = NULL){
   # Aggregate overall stats
   overall_stats <- if(has_class_probs)
     c(CM$overall, logLoss = as.numeric(lloss), AUC = roc) else CM$overall
-  if (length(levels(data[, "pred"])) > 2)
-    names(overall_stats)[names(overall_stats) == "AUC"] <- "Mean_AUC"
 
 
   # Combine overall with class-wise stats and remove some stats we don't want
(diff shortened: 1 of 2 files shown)


JackStat:
Yes sir.  https://github.com/topepo/caret/pull/596

hadjipantelis:
Oops, I did not see your pull on time and submitted a much more simplistic fix along with a new model. Thanks for sorting that out! I will close the relevant issue ( #587 ).

topepo:
Just getting to this... 

In this code: 
```r
prob_stats <-
  lapply(levels(data[, "pred"]),
         function(x){
           #Grab one-vs-all data for the class
           obs  <- ifelse(data[,"obs"] == x, 1, 0)
           prob <- data[,x]
           AUCs <- try(ModelMetrics::auc(obs, data[,x]), silent = TRUE)
           return(AUCs)
         })
```
are you sure that the AUC is using the write direction (meaning that if the AUC should be .80 and not reported to be .20)?

JackStat:
I am quite sure that this chunk resolves that issue

```
#Grab one-vs-all data for the class
           obs  <- ifelse(data[,"obs"] == x, 1, 0)
```

In cases where there is no prediction the population AUC is .50. In a sample it will drop below .50 half of the time. Other packages automatically flip it so it is always greater than .50 but the user just needs to make sure that the binary outcome that the probability pertains to is a 1 and the other a 0.

topepo:
> I am quite sure that this chunk resolves that issue

Some test cases with simulated data would help verify that =]

JackStat:
Sure. I will write since up tonight.

Issue #598, opened 2017-02-15
Regularised Logistic Regression using Liblinear


hadjipantelis:
It would be nice to have regularised logistic regression using  [LiblineaR](https://cran.r-project.org/web/packages/LiblineaR/index.html); it naturally allows L1 and L2 regularisation. It is also one of the standard implementation on [sklearn's LogisticRegression module](http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html) so that would make comparison between R and Python results a bit more straightforward.

PR #599, opened 2017-02-15
Simple `summaryFunction` crash fix and addition of regularised logistic regression through Liblinear

hadjipantelis:
The pull request has two parts:
1. In the case of using `multiClassSummary` with binary data it will automatically call the `twoClassSummary` so it does not crash any more ( #587 ).
2. It adds the regularised logistic regression models available in [LiblineaR](https://cran.r-project.org/web/packages/LiblineaR/index.html) so two standard implementation of regularised logistic regression are immediately available ( #598 ).


hadjipantelis:
Hm. I just saw that @JackStat has offered a much more appropriate fix to the crash of `multiClassSummary`. Maybe the easiest thing will be to first merge this pull first so you get the `regLogistic` model and then merge Jack's pull to overwrite the minor change my pull makes on `multiClassSummary`?

hadjipantelis:
I will close this request and re-submit it in the new few days with just the regularised logistic regression models.

PR #600, opened 2017-02-16
Addition of regularised logistic regression through Liblinear

hadjipantelis:
It adds the regularised logistic regression models available in [LiblineaR](https://cran.r-project.org/web/packages/LiblineaR/index.html). We allow for L1, L2_primal and L2_dual types of regularisation. These standard implementations of regularised logistic regression are now available ( #598 ).

### DIFF
--- a/pkg/caret/R/postResample.R
+++ b/pkg/caret/R/postResample.R
@@ -170,30 +170,41 @@ mnLogLoss <- function(data, lev = NULL, model = NULL){
   probs <- as.matrix(dataComplete[, lev, drop = FALSE])
 
   inds <- match(dataComplete$obs, colnames(probs))
-  c(logLoss = ModelMetrics::mlogLoss(dataComplete$obs, probs))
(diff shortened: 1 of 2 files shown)


topepo:
For the change to `postResample.R`, can you check the levels of `data$obs` (instead of `pred`) and use `stop` instead of issuing a warning? 

Also, please discard the change to `train.default.R`. That object _can_ be a list (based on some changes in the last year). 

Also also, please add a regression test file. You can based it on [this one](https://github.com/topepo/caret/blob/master/RegressionTests/Code/sparseLDA.R) or another from that path. Basically, these are run on every release (or significant code change) to make sure that we get [consistent results](https://github.com/topepo/caret/blob/master/RegressionTests/test_results/sparseLDA.md). 

Otherwise it looks great. Thanks!

hadjipantelis:
Thank you for the feedback and sorry for the late reply. I will look this up during the weekend. (Probably I will sync my branch with the current master first and then add the relevant files into a new pull.)

PR #608, opened 2017-03-04
Adding a couple tests

JackStat:
@topepo are there functions that you feel need to be focused on first for tests?  Not the most glorious pull request but I want to get the coverage up a bit

### DIFF
--- a/pkg/caret/tests/testthat/test_classDist.R
+++ b/pkg/caret/tests/testthat/test_classDist.R
@@ -1,0 +1,99 @@
+context('classDist')
+
+test_that("errors working", {
+
+  trainSet = 1:3
+
+  expect_error(
+    distData <- classDist(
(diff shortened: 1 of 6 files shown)


topepo:
Thanks

hadjipantelis:
Big props for starting a proper unit-test framework. Just a point to consider: Travis CI has a 10' limit for each test file and this can lead some "biggish" tests to fail because Travis is slower than expected. You might want to consider either something like `if (Sys.getenv('TRAVIS') != 'true') { ... } ` for some of the bigger tests. Notice that for Travis effectively only `tests/testthat.R` is a test. Alternatively splitting the test files into multiple ones with something like `test_check("caret", filter='_(?!BIGTEST1)(?!LONGRUN2) etc...', perl=TRUE)` might be preferable.

PR #614, opened 2017-03-09
Add weights support to xgbTree model

schistyakov:
According to xgboost documentation we can set items weights through the xgb.DMatrix. The default value for weights in caret::train is NULL, so the default behavior will stay the same


topepo:
Can you add some new test cases to the [regression tests](https://github.com/topepo/caret/blob/master/RegressionTests/Code/xgbTree.R) and also add a note in the [news file](https://github.com/topepo/caret/blob/master/pkg/caret/inst/NEWS.Rd)?

schistyakov:
OK, will do it

topepo:
Thanks

PR #658, opened 2017-05-16
Fixing error message

dfalbel:
The error message shows all variables that are in the newdata instead of the are not in newdata! :)


topepo:
Thanks!

PR #661, opened 2017-05-23
Switch to non-formula interface in ranger

mnwright:
Should be faster and more memory efficient


topepo:
Thanks for the PR. Can you resync and add a note to the NEWS.Rd file about this?

mnwright:
Done. Should the version be increased?

topepo:
Thanks!

topepo:
Nevermind. It was trivial enough to do with the web editor.

PR #679, opened 2017-07-03
Fix `preProcess` for simple data

mkobos:
`preProcess` doesn't work properly when the transformations in `preProcess` are related to individual columns only (e.g. when `method` list is equal to `list(center="y", ignore="x")` inside this function) . This is because the piece of code in `preProcess.R` file (lines 279-285) transforms the `method` list into a vector in such case and this change of type is not what further operations in the code expect. As a result, execution of line 357 ends up with error `$ operator is invalid for atomic vectors` when test in file `test_preProcess_internals.R` is run.

This pull request fixes this problem by disallowing the implicit change of the type.


topepo:
Thanks! Can you add a note in the NEWS.Rd file to reflect the fix?

mkobos:
Sure.

topepo:
Can you do a quick sync and resolve the conflict in the news file?

topepo:
Thanks again.

PR #689, opened 2017-07-13
Move optimism bootstrap routine to its own function

asardaes:
Here is the adjusted optimism bootstrap code for `nominalTrainWorkflow`. I did it on the master branch so that it is independent of the recipe integration code. In my limited tests, it does seem to give the same results as before.


topepo:
Excellent. Thanks

Issue #637, opened 2017-04-22
Possible bug in the calculation of mnLogLoss


hadjipantelis:
Hello, 

I suspect that the value returned by `mnLogLoss` is wrong. As first commented on the commit 4c9cd0eb8c1a44cd22222169883d0f9eeed5adac it appears that `caret` casts a factor vector to a numeric vector. This results in a numeric vector containing values `1`and `2` instead of `0` and `1`. These values are what is used for `y` for the calculation of the log-Loss by `ModelMetrics::logLoss` while by definition `y` is supposed to be zero or one.
In the issue https://github.com/topepo/caret/issues/107 @zachmayer  appears to correctly use something like: `obs  <- ifelse(data[,  "obs"] == class, 1, 0)` before using `Model::logLoss` but this is not present in the current release. (I use/refer at the latest (~01:30, 22 April 2017 (BST)) Github version following commit d0849e9104a8b98d855d06804e816e02dffe11bb.)

The following code snippet showcases this behaviour: 

```{R}
rm(list=ls())
devtools::load_all('git_projects/original-caret/pkg/caret/');
set.seed(1); N = 100
datTr <- twoClassSim(N)
X <- datTr[,1:(ncol(datTr)-1)]
y <- datTr[["Class"]]

datTst <- twoClassSim(N)
XTst <- datTst[,1:(ncol(datTst)-1)]
yTst <- datTst[["Class"]]

set.seed(3)
mod  <- train(X, y,
  trControl=trainControl( method="cv", number=5, 
  classProbs=TRUE, summaryFunction=mnLogLoss, savePredictions="final"))
test_pred <- predict(mod, XTst, type = "prob")
test_pred$obs <- yTst
test_pred$pred <- predict(mod, XTst)

mnLogLoss(test_pred, lev = levels(test_pred$obs)) # 0.4129...

# Reproducing outside mnLogLoss/postResample 
ModelMetrics::logLoss(predicted = test_pred$Class1, 
                         actual =  as.numeric(test_pred$obs) ) # 0.4129...

obs  <- ifelse(test_pred$obs == 'Class1', 1, 0)
ModelMetrics::logLoss(predicted = test_pred$Class1, actual =  obs ) # 0.5738...
```

### Session Info:
```{R}
   R version 3.3.3 (2017-03-06)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Debian GNU/Linux 8 (jessie)

locale:
[...]

PR #696, opened 2017-07-19
Fix issue #637: incorrect two-class mnLogLoss

sergeykorop:
Fix mnLogLoss to work properly for two-class input.

Add unit test for this particular case.

### DIFF
--- a/pkg/caret/R/misc.R
+++ b/pkg/caret/R/misc.R
@@ -585,3 +585,97 @@ subset_x <- function(x, ind) {
     x
 }
 
+optimismBoot <- function(ctrl, x, y, wts, iter, lev, method, mod, predicted, submod, loop) {
+  indexExtra <- ctrl$indexExtra[[iter]]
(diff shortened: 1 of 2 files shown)


sergeykorop:
As an additional improvement, the entire if-statement
```r
  if(nlevels(dataComplete$obs) == 2){
    logLoss <- ModelMetrics::logLoss(dataComplete$obs == lev[2], probs[, lev[2]])
  } else {
    logLoss <- ModelMetrics::mlogLoss(dataComplete$obs, probs)
  }
```
can be replaced by it's else branch:
```r
logLoss <- ModelMetrics::mlogLoss(dataComplete$obs, probs)
```
To the best of my knowledge, they should be equivalent except, maybe, for some difference in floating-point accuracy and/or performance.

topepo:
@JackStat Can you do a quick review of this? It is modifying your contribution so you might be interested.

PR #701, opened 2017-07-29
Use requireNamespace instead of require

lepennec:
Dear Max,

I've finally made it. Here is a pull request in which each require has been replaced by requireNamespace and each model has been modified accordingly. It pass all the tests except the ones for elm, GFS.GCCL, GFS.Thrift and ranger. Noe that I had to fix a typo in adaptive.R...

Please feel free to comment. I can rewrite the commits if you find them to monolitic...

Erwan


topepo:
I'm testing...

`gamboost.R` has the line
```r
import::from(mboost, bbs, .into = "mboost")
```
Should `import` be added to the `$library` vector?

lepennec:
Indeed! I have modified the two models in which I had to use import.

topepo:
Thanks!

PR #702, opened 2017-08-01
Fix issue #637: incorrect two-class mnLogLoss

sergeykorop:
Fix mnLogLoss to work properly for two-class input.

Add unit test for this particular case.

Supersedes PR #696

### DIFF
--- a/pkg/caret/R/postResample.R
+++ b/pkg/caret/R/postResample.R
@@ -173,7 +173,7 @@ mnLogLoss <- function(data, lev = NULL, model = NULL){
 
   inds <- match(dataComplete$obs, colnames(probs))
   if(nlevels(dataComplete$obs) == 2){
-    logLoss <- ModelMetrics::logLoss(dataComplete$obs, probs)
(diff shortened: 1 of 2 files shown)


sergeykorop:
As an additional improvement, the entire if-statement

```r
  if(nlevels(dataComplete$obs) == 2){
    logLoss <- ModelMetrics::logLoss(dataComplete$obs == lev[2], probs[, lev[2]])
  } else {
    logLoss <- ModelMetrics::mlogLoss(dataComplete$obs, probs)
  }
```
can be replaced by it's else branch:

```r
logLoss <- ModelMetrics::mlogLoss(dataComplete$obs, probs)
```

To the best of my knowledge, they should be equivalent except, maybe, for some difference in floating-point accuracy and/or performance.

sergeykorop:
@JackStat My bug fix proposal is related to your code. I would be very thankful if you review it, please.

topepo:
Thanks. I agree on simplifying too.

sergeykorop:
@topepo Thank you for cooperation!

PR #705, opened 2017-08-05
Change reference to undefined variable 'mod' in thresholder

jameslamb:
`R CMD check` throws the following note:

```
* checking R code for possible problems ... NOTE
thresholder: no visible binding for global variable ‘mod’
Undefined global functions or variables:
  mod
```

Looking at the function signature, it appears that `thresholder` should be referencing input argument `x`, not `mod`.


topepo:
Thanks

PR #718, opened 2017-08-21
Describe parameters to defaultSummary

DarioS:


topepo:
`data` is constrained to data frames only (see a few lines down with `data$obs`). I'll change it once the PR is in sine I should add some more info about recipes too. 

Thanks

PR #722, opened 2017-08-22
mxnetAdam model and test

hadjipantelis:
mxnet using Adam optimiser.  Model and regression test files. Model not parsed.

### DIFF
--- a/RegressionTests/Code/mxnetAdam.R
+++ b/RegressionTests/Code/mxnetAdam.R
@@ -1,0 +1,160 @@
+timestamp <- Sys.time()
+library(caret)
+library(plyr)
+library(recipes)
+library(dplyr)
+
+modelZ <- "mxnetAdam"
+
+## In case the package or one of its dependencies uses random numbers
+## on startup so we'll pre-load the required libraries: 
+
+for(i in getModelInfo(modelZ)[[1]]$library)
+  do.call("requireNamespace", list(package = i))
+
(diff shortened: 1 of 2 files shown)


topepo:
```r
out <- data.frame(nlayers = sample(2:20, replace = TRUE, size = len), 
                  nnodes =  sample(c(2:99), replace = TRUE, size = len), 
```
Wow, that's a $^@*-ton of parameters. 

Here is what the other `mxnet` model does:
```r
grid = function(x, y, len = NULL, search = "grid") {
  if(search == "grid") {
    out <- expand.grid(layer1 = ((1:len) * 2) - 1, layer2 = 0, layer3 = 0,
                       learning.rate = 2e-6, 
                       momentum = 0.9, 
                       dropout = seq(0, .7, length = len))
  } else {
    out <- data.frame(layer1 = sample(2:20, replace = TRUE, size = len),
                      layer2 = 0,
                      layer3 = 0,
                      learning.rate = runif(len),
                      momentum = runif(len, min = .9),
                      dropout = runif(len, max = .7))
  }
  out
},
```
If you want a relatively small model, then it isn't an issue. Otherwise, sampling `layer2` and `layer3` can generate a pretty complex model.

hadjipantelis:
OK. Saw it. I will go ahead and amend this!

topepo:
Thanks!

Issue #729, opened 2017-09-01
Inactive unit tests


sergeykorop:
Some unit test files in [caret/pkg/caret/tests/testthat](https://github.com/topepo/caret/tree/master/pkg/caret/tests/testthat) have names without `test_` prefixes which are [required](http://r-pkgs.had.co.nz/tests.html) by `testthat` framework:

> Test structure
> A test file lives in tests/testthat/. Its name must start with test.

As a consequence, such files (e.g. [preProcess_methods.R](https://github.com/topepo/caret/blob/master/pkg/caret/tests/testthat/preProcess_methods.R)) are skipped when running the test suite with `devtools::test`.

topepo:
Thanks. Made the change.

PR #730, opened 2017-09-02
Add `rangeBounds` parameter to preProcess()

sergeykorop:
This PR proposes following improvement: by adding new parameter, `rangeBounds` to `preProcess()` we will be able to specify an explicit interval for `range` transformation, e.g. `[-0.5, 0.5]` instead of `[0, 1]` used by default.

### DIFF
--- a/pkg/caret/R/preProcess.R
+++ b/pkg/caret/R/preProcess.R
@@ -41,7 +41,7 @@ invHyperbolicSineFunc <- function(x) log(x+sqrt(x^2+1))
 #' from the data in \code{x}) from the predictor values while \code{method =
 #' "scale"} divides by the standard deviation.
 #' 
-#' The "range" transformation scales the data to be within [0, 1]. If new
+#' The "range" transformation scales the data to be within \code{rangeBounds}. If new
 #' samples have values larger or smaller than those in the training set, values
 #' will be outside of this range.
 #' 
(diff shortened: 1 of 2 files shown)


sergeykorop:
This PR interferes with issue #729: unit tests for `range` transformation of `preProcess()` are in [preProcess_methods.R](https://github.com/topepo/caret/blob/master/pkg/caret/tests/testthat/preProcess_methods.R) which is not active for now.

To work-around this issue, I reproduced existing tests related to `range` in new test file, `test_preProcess_range.R` and added some new ones covering new functionality. If issue #729 will be settled with bringing `preProcess_methods.R` back to a pool of active tests, these new tests can be re-integrated back to that file.

topepo:
Thanks for adding this. You could also get this effect by using the `recipes` interface to `train` (just as an fyi)

topepo:
@sergeykorop Based on the reverse dependency checks, can you make some quick changes to ensure that the new version of range scaling works with objects from previous versions where those objects don't exist? 

For example, the printing is off:
```r
> standardObj
Created from 214 samples and 2343 variables

Pre-processing:
  - ignored (0)
  - re-scaling to [, ] (2343)
```
Also, for old objects, these lines fail:
```r
newdata[, object$method$range] <- sweep(newdata[, object$method$range, 
    drop = FALSE], 2, (object$ranges[2, ] - object$ranges[1, 
    ])/(object$rangeBounds[2] - object$rangeBounds[1]), "/")
newdata[, object$method$range] <- sweep(newdata[, object$method$range, 
    drop = FALSE], 2, object$rangeBounds[1], "+")
```

sergeykorop:
@topepo See PR #790 for bug fix.

PR #741, opened 2017-09-26
remove unnecessary space

smsaladi:
tiny suggestion, but because there is a surrounding `paste`, this space is unnecessary. Currently, the output is formatted as follows:

`pearson was used to select the optimal model using  the largest value.`

image
(not sure it shows clearly as text above)


topepo:
Thanks!

PR #746, opened 2017-10-11
Prefix predict for every kernlab use

lepennec:
Dear Max,

it seems that the predict function has to be explicitly prefixed when using kernlab otherwise NAs are prediicted...

Erwan

### DIFF
--- a/models/files/gaussprLinear.R
+++ b/models/files/gaussprLinear.R
@@ -12,12 +12,12 @@ modelInfo <- list(label = "Gaussian Process",
 				     kpar = list(), ...)
                     },
                   predict = function(modelFit, newdata, submodels = NULL) {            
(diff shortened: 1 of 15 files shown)


topepo:
Huh. That's odd. 

It's a good idea to do this anyway but do you can an example that I can test with?

Also, should they be class specific (e.g. `kernlab::predict.ksvm` instead of `kernlab::predict`)?

lepennec:
It looks like kernlab::predict is a generic method so that specifying the
class is useless. Let me wrap up an example tomorrow...

lepennec:
```
library(caret)
model <- train(Species ~ ., data = iris, method = "svmRadial")
predict(model)
```

topepo:
Thanks

Issue #743, opened 2017-09-28
doAzureParallel: could not find function "subset_x"


brnleehng:
Hi,

I'm trying to run doAzureParallel as my foreach backend. I'm having issues with not finding 'subset_x' in the caret package. 

Will I need to do following as the PR below with caret:::subset_x?
https://github.com/topepo/caret/pull/500

Error:
loaded caret and set parent environment
Warning messages:
1: In eval(expr, envir, enclos) :
  model fit failed for Fold1.Rep1: mtry=1 Error in rownames(x) : could not find function "subset_x"

2: In eval(expr, envir, enclos) :
  model fit failed for Fold1.Rep1: mtry=5 Error in rownames(x) : could not find function "subset_x"

3: In eval(expr, envir, enclos) :
  model fit failed for Fold2.Rep1: mtry=1 Error in rownames(x) : could not find function "subset_x"

4: In eval(expr, envir, enclos) :
  model fit failed for Fold2.Rep1: mtry=5 Error in rownames(x) : could not find function "subset_x"

5: In eval(expr, envir, enclos) :
  model fit failed for Fold1.Rep2: mtry=1 Error in rownames(x) : could not find function "subset_x"

### Session Info:
[1] "argsList"           "chunkSize"          "cloudCombine"      
[4] "enableCloudCombine" "exportenv"          "expr"              
[7] "packages"           "pkgName"           
[1] "caret"
R version 3.3.2 (2016-10-31)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: CentOS Linux 7 (Core)

locale:
[1] C

attached base packages:
[1] methods   stats     graphics  grDevices utils     datasets  base     

other attached packages:
[1] randomForest_4.6-12 caret_6.0-71        ggplot2_2.2.0      
[4] lattice_0.20-34    

loaded via a namespace (and not attached):
 [1] Rcpp_0.12.7            nloptr_1.0.4           plyr_1.8.4            
 [4] iterators_1.0.8        tools_3.3.2            lme4_1.1-12           
 [7] jsonlite_1.1           tibble_1.2             nlme_3.1-128          
[10] gtable_0.2.0           mgcv_1.8-15            Matrix_1.2-7.1        
[13] foreach_1.4.3          CompatibilityAPI_1.1.0 parallel_3.3.2        
[16] SparseM_1.74           stringr_1.0.0          MatrixModels_0.4-1    
[19] stats4_3.3.2           grid_3.3.2             nnet_7.3-12           
[...]

PR #748, opened 2017-10-12
Add missing exports required for some foreach backend

lepennec:
I've stumbled upon a bug for some foreach backend (doFuture with multisession for instance...). There are some missing exports in  workflows.R. It seems that completing the list as proposed below fix this issue.

### DIFF
--- a/pkg/caret/R/workflows.R
+++ b/pkg/caret/R/workflows.R
@@ -75,12 +75,10 @@ nominalTrainWorkflow <- function(x, y, wts, info, method, ppOpts, ctrl, lev, tes
   keep_pred <- isTRUE(ctrl$savePredictions) || ctrl$savePredictions %in% c("all", "final")
   pkgs <- c("methods", "caret")
   if(!is.null(method$library)) pkgs <- c(pkgs, method$library)
-  export <- c("optimism_xy", "ctrl", "requireNamespaceQuietStop",
-              "x", "y", "wts", "info", "method", "ppOpts", "ctrl", "lev",
-              "testing","pkgs", "resampleIndex")
+  export <- c()
 
-  result <- foreach(iter = seq(along = resampleIndex), .combine = "c", .verbose = FALSE, .export = export) %:%
-    foreach(parm = 1L:nrow(info$loop), .combine = "c", .verbose = FALSE, .export = export)  %op%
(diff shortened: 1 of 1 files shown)


freezby:
I had the same problem using doMPI on a cluster of two nodes (localhost and one remote computer)
But to get things working I had to complete the export variable in a different way : 
  export <- c("optimism_xy", "requireNamespaceQuietStop","subset_x","pred_failed",
              "model_failed", "fail_warning", "fill_failed_pred", "fill_failed_prob",
              "trim_values", "outcome_conversion", "flatTable", "ppOpts", "createModel",
              "predictionFunction","probFunction"
              )

lepennec:
OK. It looks like my list should be enlarged...

brnleehng:
I had a similar issue with doAzureParallel, it seems that all of the internal functions will need to be 
in the export list since it depends on what caret function you are using 

#743

brnleehng:
I ran one of our samples codes. Installed your branch, It works for me. 

library(DAAG)
library(caret)

set.seed(998)
inTraining <- createDataPartition(spam7$yesno, p = .75, list = FALSE)
training <- spam7[ inTraining,]
testing  <- spam7[-inTraining,]

fitControl <- trainControl(## 10-fold cross validation
                           method = "repeatedcv",
                           number = 10,
                           ## repeat 10 times
                           repeats = 10,
                           classProbs = TRUE,
                           summaryFunction = twoClassSummary,
                           search = "random",
                           ## run on the parallel backend
                           allowParallel = TRUE)

rf_fit <- train(
                 yesno ~ .,
                 data = training,
                 method = "rf",
                 metric = "ROC",
                 tuneLength = 30,
                 trControl = fitControl)

rf_fit

topepo:
Thanks for fixing this. I run a bunch of tests using `doMC`. I'll add some more for `doParallel`. The issue is somewhat OS dependent (and cluster type dependent) too.

topepo:
@brnleehng have you tested these updates on `doAzureParallel`? I'm getting a release together and it would be good to have confirmation. 

@lepennec Anything else that you think needs to happen?

lepennec:
I expect that if we cover sufficient different backends, which seems to be the case, we should not have more surprise...

PR #756, opened 2017-10-19
Minor fixes to mxnet* model files.

hadjipantelis:
Very minor adjustments to `mxnet*` related code. 

Harding-coding verbose as `FALSE` will probably increase convenience. 
Similarly adding `packagename::functionname()` syntax for `mx.set.seed`; I forgot at the time.

### DIFF
--- a/models/files/mxnet.R
+++ b/models/files/mxnet.R
@@ -45,8 +45,8 @@ modelInfo <- list(label = "Neural Network",
                                            eval.metric = mx.metric.rmse, 
                                            array.layout = "rowmajor",
                                            activation = rep( as.character(param$activation), length(num_units)),
+                                           # Use He/MSRA when available in R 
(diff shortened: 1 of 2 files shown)


topepo:
While I agree that `verbose = TRUE` sucks as a default (I tried to talk them out of it and even submitted a PR), there is no way to get verboseness with these changes. You could intercept the `...` and change it there if it isn't passed in, but then it triggers a bunch of `do.call` nonsense. Also, this would be inconsistent with a lot of other methods (`gbm`, `nnet` and the `keras` models, etc.). 

Can you remove the `verbose` change. Everything else looks good.

hadjipantelis:
No problem. Done. I removed `verbose` from `mxnetAdam` to ensure common interface.

I added a comment about the initialiser choice in the fuure: Given `relu` activation is used by default in the case of a grid search, "ideally" we would like to use He initialisation instead of Xavier (See https://arxiv.org/pdf/1502.01852.pdf) but that is not a huge delta at this point.

topepo:
Thanks

PR #774, opened 2017-11-11
Remove duplicate unit tests introduced in PR #730

sergeykorop:
Separate unit tests for preProcess/range method added in PR #730
are no longer needed after fixing issue #729. Their functionality
has been moved to generic preProcess/methods suite.


topepo:
Thanks

PR #782, opened 2017-11-17
Avoid leaving option() changed when model.matrix fails

r-suzuki:
In some situations `x <- model.matrix(Terms, m)` fails and `options()$constrasts` leaves changed to `"contr.ltfr"`. Calling `options(contrasts = oldContr)` via `on.exit()` solves this problem.

A reproducible example is shown below. The same thing happens when a user loads `dummyVars` object from .RData file and calls `predict` function before loading the package.

```r
options()["contrasts"] # Normally the first element is "contr.treatment"
unloadNamespace("caret") # Unload the package

dummies <- caret::dummyVars(Sepal.Length ~ ., data = iris)
head(predict(dummies, newdata = iris)) # It fails because the package is not loaded

options()["contrasts"] # The first element has been changed to "contr.ltfr"

model <- lm(Sepal.Length ~ ., data = iris) # It also fails because the "contrasts" option has been changed
```

### DIFF
--- a/pkg/caret/R/dummyVar.R
+++ b/pkg/caret/R/dummyVar.R
@@ -221,11 +221,11 @@ predict.dummyVars <- function(object, newdata, na.action = na.pass, ...)
     newContr <- oldContr
     newContr["unordered"] <- "contr.ltfr"
     options(contrasts = newContr)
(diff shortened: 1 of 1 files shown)


topepo:
Looks good. Thanks

PR #788, opened 2017-11-21
Added feature to show training progress

washcycle:
I think this is pretty awesome. Uses the resampleIndex and counts how many sample index vectors have been trained and uses that as a metric for training progress. It is compatible with parallel processing.

My primary driver for this was to display the progress of training a caret model in an R Shiny app. Works amazingly well for my purposes.

I tried to make it robust as possible to errors and disabled it by default.

_I think this is worthy of a beer at the RStudio conference in January ;)_

**Sample usage:**

    ctrl <- caret::trainControl(method = "repeatedcv", showProgress=TRUE)
    model <- caret::train(Species ~ ., data = iris, method = "rf", trControl = ctrl)


topepo:
Thanks! I've resisted the requests to do this for a while but since someone else did the work...

It might take me a bit to look at this and decide on what to do. A few questions off of the top of my head:

* does this play well with `verboseIter` and what (besides coolness) does this offer that it does not?
* for the submodel trick models, how does this determine the denominator? We might evaluate 100 models but only fit 10.

washcycle:
1. Besides coolness it works with R Shiny that's the real reason I added it. 
2. Good question on the verboseIter, I don't know if it will play well. It only displays a progress bar by default. Which might look funky if both are printing at the same time. Might have to give one of the settings priority and disable the other.
3. I'd have to look at the submodels, not sure.

washcycle:
Another thing that came to me this morning is that since the combine function of foreach overloaded, it is extensible. Instead of a progress bar in the console it could be a plot that updates.

topepo:
> it could be a plot that updates.

😱 

Maybe I'm too old school, but that sounds horrifying to me. I've been pleading for `keras` to not do that.

washcycle:
I don't know if you've have a chance to look at this, but I am not too familiar how the logic of the workflows. I did some digging and it looks like it would be prudent to implement this for all the workflows.. potentially. If you think that is a good idea I can do that.

PR #790, opened 2017-11-22
Handle `rangeBounds` for old `preProcess()` outputs

sergeykorop:
Preprocessing context objects created by past versions of `preProcess()`
don't contain interval to be used for `range` transform and we have to
handle them in a special way assuming [0, 1] interval by default.


topepo:
Excellent. Thanks

sergeykorop:
@topepo Thank you, too, for discovering this problem.

PR #812, opened 2018-01-07
Fixed xgbDART bug

hadjipantelis:
In the case of using `dart` booster, `predict(xgboostmodel, newdata)` does not use all trees by default, as hinted in the `predict.xgb.Booster`. It appears that it picks at random (?) one iteration and returns the relevant results; we need to fix `predict` such that `predict(xgboostmodel, newdata, ntreelimit = modelFit$niter)`. This rendered the training procedure correct up until the last iteration (which was actually done first) in `predict`). The current `RegressionTests` code was able to pass (probably by sheer bad luck). I wrote a small additional test that should be much more likely to fail if this unwanted behaviour reappears. 

(I will create a relevant documentation fix request to `xgboost` - I have no idea why this behaviour is like that, it breaks compatibility with the `tree` as well as the `glm` booster...)


hadjipantelis:
issue reported in: https://github.com/dmlc/xgboost/issues/3015.

topepo:
I can't resolve the conflict on `models.RData`. Can you sync your PR to the master branch and then re-run `parseModels.R`?

hadjipantelis:
No problem. Done.

topepo:
Thanks

PR #859, opened 2018-03-15
Fix misquotation of meta tag of GitHub repository

erictleung:
The live site has a stray link in the top left corner of the page. This fixes the quotation of the link that should have single quotes rather than double quotes.

**Screenshot**:

Before:

image

After:

image


topepo:
Thanks

PR #879, opened 2018-05-08
Added models file for tuning metaforest models

cjvanlissa:
MetaForest is a random forests based approach to meta-analysis. The package uses ranger, which is alreasdy included in the models list, but it has one additional tuning parameter, and requires different S3 methods. I would like to allow users to use caret to train and tune these models, and hope you will consider accepting this models file to facilitate that! I am happy to implement any desired changes.

Sincerely,
Caspar

### DIFF
--- a/models/files/metaforest.R
+++ b/models/files/metaforest.R
@@ -10,7 +10,7 @@ modelInfo <- list(label = "MetaForest",
                   },
                   loop = NULL,
                   type = "Regression",
-                  parameters = data.frame(parameter = c("whichweights", "mtry", "min.node.size"), # Add random, unif, fixed
+                  parameters = data.frame(parameter = c("whichweights", "mtry", "min.node.size"),
                                           class = c("character", "numeric", "numeric"),
                                           label = c("Type of weights to use", "#Randomly Selected Predictors",
                                                     "Minimal Node Size")),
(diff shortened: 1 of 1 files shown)


topepo:
Thanks for submitting this. 

I don't think that I want to include it. Since the user needs to pass in the `vi` parameter (or is constrained to have a variable with that name) it will cause issues down the line that can't be programmatically addressed. Also, the lack of formula and recipe methods would be a similar (and bigger) issue.  You've documented these issues well, but since `caret` doesn't easily handle non-predictor variables when fitting the model, it is a square peg in a round hole. 

(Model tuning for this will be easier using `rsample` and `recipes` but that modeling system isn't quite finished yet. I'd be happy to mock up and example if that helps.)

topepo:
> I would also be happy to adapt the interface of the MetaForest and ClusterMF functions to ensure compatibility with caret.

Sorry to confuse; the issue is with `caret` not being able to handle variables that are not predictors or outcomes in its fitting module. 

That was one of the motivations for `recipes`; you can declare roles for some variables (like the `vi` argument in your function) and these variables can be preprocessed and resampled like everything else. 

The only suggestion that I can make to you is to have 

* the formula method for `MetaForest` treat the argument for the variances of the effect sizes as data and not the variable name similar to how most models (like `lm`) treat `weights`.
* have a basic matrix method that has main arguments for X, y, and vi. then the formula method could pass its parse/processed arguments to this default method. For an example, see `MASS:::lda.formula` and `MASS:::lda.default`.

cjvanlissa:
Thank you for these helpful suggestions. I will implement this when I have more time.

PR #881, opened 2018-05-09
Request to add model 'pre' for package 'pre'

marjoleinF:
Dear caret team,

Thank you for your work on the caret package! 

I've set up a model list for my package for caret, perhaps I may suggest it hereby for inclusion in caret? The package is 'pre', which is short for prediction rule ensembles. Has been on CRAN for some time now ([https://CRAN.R-project.org/package=pre](https://CRAN.R-project.org/package=pre)).

The code below provides some tests:

````
## Load packages:
library("caret")
# install.packages("pre") # install if needed
library("pre")

## source model list:
source(url("https://raw.githubusercontent.com/marjoleinF/caret/master/models/files/pre.R"))
closeAllConnections()

## load data:
airq <- airquality[complete.cases(airquality),]
y <- airq$Ozone
x <- airq[,-1] 

## With default settings and 'ntrees = 5' to reduce computation time:
set.seed(42)
prefit1 <- train(x = x, y = y, method = modelInfo, 
                 trControl = trainControl(number = 1), 
                 tuneGrid = modelInfo$grid(ntrees = 5))
prefit1

## With larger tuneGrid:
set.seed(42)
tuneGrid <- modelInfo$grid(x = x, y = y, type = c("linear", "rules", "both"),
                           maxdepth = 2:4, ntrees = 5)
prefit2 <- train(x = x, y = y, method = modelInfo, 
                 trControl = trainControl(number = 1),
                 tuneGrid = tuneGrid)
prefit2

## With binary response:
y_bin <- factor(y > mean(y))
levels(y_bin) <- c("no", "yes")
set.seed(42)
prefit3 <- train(x = x, y = y_bin, method = modelInfo, 
                 tuneGrid = modelInfo$grid(ntrees = 5),
                 trControl = trainControl(number = 1, classProbs = TRUE), 
                 family = "binomial")
prefit3
predict(prefit3, type = "prob")[1:6,]

## Model Ozone as count response:
set.seed(42)
prefit4 <- train(x = x, y = y, method = modelInfo, 
                 tuneGrid = modelInfo$grid(ntrees = 5),
                 trControl = trainControl(number = 1), 
                 family = "poisson")
prefit4

## With multinomial response:
y_multin <- y_multin <- factor(cut(y, breaks = c(0,20,50, 200)))
levels(y_multin) <- c("low", "medium", "high")
set.seed(42)
prefit5 <- train(x = x, y = y_multin, method = modelInfo, 
                 tuneGrid = modelInfo$grid(ntrees = 5),
                 trControl = trainControl(number = 1, classProbs = TRUE), 
                 family = "multinomial")
prefit5
predict(prefit5, newdata = x , type = "prob")[1:6,]
```


topepo:
I'd like to inlcude it but there are some things that need to be changed:

1. it looks like there is an importance method, so adding a `varImp` module would be good.

1. Please make the labels for the tuning parameters more readable and consistent with the other methods (e.g. for `mtry`, `ntrees `, etc).

1. Add regression tests. There are a lot of examples in the repo already.

1. The `grid` module should also have a path for random search. Also, there are too many arguments to the `grid` function. Although the extras will be ignored, it is better to eliminate them. 

1. In the `fit` module, I would avoid using `do.call` unless you need to intercept one or more arguments from `...` for modification. The call can end up duplicating the data set when this is used. Also, consider reducing the risk of duplicate names in `data` by using something other than "response". Other modules use `.outcome`

1. Consider setting `family` based on the type of data in `y` so the user doesn't have to. 

1. If the model cannot deal with 3+ classes, put a check in the code for that somewhere and add the tag "Two Class Only"

1. The `sort` module should be able to order the models from least complex to most complex based on their parameter values. That should be filled in. It's a good idea to add a `levels` module too. 


One other note about the package. You might consider adding a default method that has arguments similar to `X` and `y` and then have the formula method call that. See `lda.default` and `lda.formula` in the MASS package for an example. The formula method can [waste a lot of time when there are a large number of predictors](https://rviews.rstudio.com/2017/03/01/the-r-formula-method-the-bad-parts/).

topepo:
One other question (unrelated to this PR): is there a reference for the interaction test used here? We're writing a book chapter on this topic and that would be good to read.

marjoleinF:
Thanks for the swift reply and helpful comments! Will implement and recommit.

Three questions:

1. There is an importance method, but it does not add up to 100 as in other methods. Better to rescale to 100 (but then results don't match with importance function in package itself, but can include note about that) or keep original scaling?

4. Indeed, lots of parameters to tune. To be sure, better to make an educated guess on which ones are most relevant and only implement those in grid? 

8. With the many parameters, it is difficult to predict which model will be most complex. The best and simplest ordering could be obtained by counting the number of terms and variables in the final model. Can the sort function use the fitted model instead of the tuning parameter values to sort?

Furthermore, on 6, family is set by default already (I often add it just to be sure); on 7, pre takes any number of classes, test was just to illustrate it works with > 2, so this should create no problems.

Thanks for the note on formulas vs x and y, very enlightening.

Reference for the interaction tests is Section 8 of Friedman & Popescu (2008). Basically tests the extent to which model predictions differ from those of an additive model. No further research has been done an it, as far as I know. 

Friedman, J. H., & Popescu, B. E. (2008). Predictive learning via rule ensembles. The Annals of Applied Statistics, 2(3), 916-954. [https://projecteuclid.org/download/pdfview_1/euclid.aoas/1223908046](https://projecteuclid.org/download/pdfview_1/euclid.aoas/1223908046)

topepo:
> There is an importance method, but it does not add up to 100 as in other methods. Better to rescale to 100 (but then results don't match with importance function in package itself, but can include note about that) or keep original scaling?

Keep the original scaling. `varImp` rescales them and there is an option to keep the original scale. 

> Indeed, lots of parameters to tune. To be sure, better to make an educated guess on which ones are most relevant and only implement those in grid?

I'd say leave them in. That what random search is really good at and I tend to use it when there are 4 or more parameters. 

> With the many parameters, it is difficult to predict which model will be most complex. 

Yeah. There is no definitive way of estimating this. For example, I have no idea how `mtry` would effect the complexity. An educated guess that uses at least one tuning parameter will be fine. 

This was implemented for models like `rpart` when `maxdepth` is used instead of `cp`. You end up with multiple tuning parameters with exactly the same performance. I wanted to pick the least complex model to finalize. 

> The best and simplest ordering could be obtained by counting the number of terms and variables in the final model. Can the sort function use the fitted model instead of the tuning parameter values to sort?

Right now it just gets the tuning parameter values. A rough sorting will work fine. 

> Furthermore, on 6, family is set by default already (I often add it just to be sure); on 7, pre takes any number of classes, test was just to illustrate it works with > 2, so this should create no problems.

👍

> Thanks for the note on formulas vs x and y, very enlightening.

Hadley and I are in the process of writing an "R modeling package manifesto" that will list ideas like this (and their genesis) in the hope that people will use them as conventions (since there really are none). 

Thanks for the reference. I read that back when it came out and had forgetting about it.

models/files/pre.R
```diff
@@ -0,0 +1,178 @@
...
+        maxdepth = sample(2L:6L, size = len, replace = TRUE), 
+        learnrate = sample(c(0.001, 0.01, 0.1), size = len, replace = TRUE),
+        mtry = sample(c(ceiling(sqrt(ncol(x))), ceiling(ncol(x)/3), ncol(x)), size = len, replace = TRUE),
+        ntrees = rep(500, times = len),
+        use.grad = sample(c(TRUE, FALSE), size = len, replace = TRUE),
+        penalty.par.val = sample(c("lambda.1se", "lambda.min"), size = len, replace = TRUE))
+    }
+    return(out)
+  },
+  fit = function(x, y, wts = NULL, param, lev = NULL, last = NULL,
```
topepo: `family` can be passed in via `...` but you should probably intercept the `...` and set `family` if it is not passed in. See [the `glm` code](https://github.com/topepo/caret/blob/1cd9e2283e48a99828e6695f0cdbcb0794b08e10/models/files/glm.R#L14) for an example. 

It's a it more work but most of the other modules do so and it would cut down on a lot of SO questions and issues filed.

topepo:
Can you add some regression tests? Lots of examples [here](https://github.com/topepo/caret/tree/master/RegressionTests/Code).

marjoleinF:
It took me some time to fine tune and fix errors. I have implemented your suggestions (thanks!) and added regression tests, by now. Please let me know if any omissions/bugs/mistakes. 

The regression tests work with the current github version of "pre", but not the current CRAN version, as the predict method yielded an error with the leave-one-out CV examples (did not allow prediction of a single observation, but this is fixed now).

topepo:
> The regression tests work with the current github version of "pre", but not the current CRAN version, as the predict method yielded an error with the leave-one-out CV examples (did not allow prediction of a single observation, but this is fixed now).

I'm getting ready to do a release in the next week. Do you know if the CRAN version of `pre` has been updated for this?

marjoleinF:
Thanks for the heads-up! Have checked against current CRAN version of pre. After small adjustments, both files in models and RegressionTests work with current R version now.

models/files/pre.R
marjoleinF: Dots are intercepted now and family argument set if not passed to dots.

PR #882, opened 2018-05-13
Add createMultiFolds() to help page

schoonees:
At present, createMultiFolds() is partly documented under ?createDataPartition (under Details mostly). But it does not show up under the Usage section.

So I propose to add 

#' @rdname createDataPartition

for this function so that it will show up there.

If this is intentional, please ignore.


topepo:
Thanks

PR #893, opened 2018-06-05
To solve the problem: when using xgbTree, the memory would not be released

Japrin:
This is xgbost's  problem, see issue [3045](https://github.com/dmlc/xgboost/issues/3045).
To solve it, just simply set mod to NULL and call gc() in the proper place of R/adaptive.R and R/workflows.R.


topepo:
Closed due to the problem appearing to be in `xgboost` (and I don't think that this implementation will help). Thanks though.

PR #911, opened 2018-07-10
Bugfix to make train work with step_string2factor()

smaakage85:
When calling 'train' with the 'recipes' interface, the options for the 'prep' call are hardcoded. Especially the user should be able to set the 'stringsAsFactors' option for the 'prep' call, when calling 'train'. Otherwise steps like 'step_string2factor()' will not work.

### DIFF
--- a/pkg/caret/R/train.default.R
+++ b/pkg/caret/R/train.default.R
@@ -960,7 +960,8 @@ train.recipe <- function(x,
                          maximize = ifelse(metric %in% c("RMSE", "logLoss", "MAE"), FALSE, TRUE),
                          trControl = trainControl(),
                          tuneGrid = NULL,
-                         tuneLength = ifelse(trControl$method == "none", 1, 3)) {
+                         tuneLength = ifelse(trControl$method == "none", 1, 3),
+                         stringsAsFactors = TRUE) {
   startTime <- proc.time()
   
   ## get a seed before packages are loaded or recipes are processed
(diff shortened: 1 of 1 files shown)


smaakage85:
Hi again,  @topepo  :) 

Well, maybe I should have put this under "Issues" to start with. I am a bit new to this. 

I guess, what I am trying to is, that there is a problem with carets support for step_string2factor(), see example below:

---

library(dplyr)
library(magrittr)
library(caret)
library(recipes)
library(tidyselect)

df <- 
  iris %>%
  mutate_at("Species", as.character)

rec <- 
  recipes::recipe(df) %>%
  add_role(Sepal.Length, new_role = "outcome") %>%
  add_role(everything(), -Sepal.Length, new_role = "predictor")

example1 <-
caret::train(
  rec,
  data = df,
  method = "lm",
  trControl = caret::trainControl(
    method = "cv",
    number = 2
  )
)

example2 <-
caret::train(
  rec %>%
    step_string2factor(Species),
  data = df,
  method = "lm",
  trControl = caret::trainControl(
    method = "cv",
    number = 2
  )
)

---
example1 works, because caret::train sets stringsToFactors to TRUE.

example2 fails (unexpectedly) with the following error message:
"Error: The following variables are not character vectors: `Species`". As I see it, this is due to the fact, that stringsAsFactors is set to TRUE.

topepo:
So it will be a factor once `train` starts using the recipe. It looks like that is the outcome that you are looking for. Although it is inconsistent, I don't want to add another option. I've added an issue to recipes to make this not fail if the variable is already a factor.

topepo:
That should do it. Please test so that I can close this.

smaakage85:
Sorry for the delay, have been on vacation for a couple of days.

Beautiful. Now it works, as I would expect it to! :)

Issue #902, opened 2018-06-21
Building custom Ordinal Forest Model in caret's train framework


HanjoStudy:
##  Needing help with
I have recently come across the [orindalForest](https://cran.r-project.org/web/packages/ordinalForest/ordinalForest.pdf) package. This package focuses on an implementation of the well known [ranger](https://cran.r-project.org/web/packages/ranger/ranger.pdf) `Random Forest` model. With this in mind, I am trying to integrate the model into the caret framework without any luck. Here is what I have so far:

## The code

Package can be installed from `CRAN`
```{R}
# install.packages("ordinalForest")

# start by naming my method to pass to train
ordinalForest <- list(type = "Classification",
              library = "ordinalForest",
              loop = NULL)

# define the tuning parameters
prm <- data.frame(parameter = c("nsets", "ntreeperdiv", "ntreefinal", "npermtrial", "nbest"),
                  class = rep("numeric", 5),
                  label = c("Score sets", "Number of Trees (small)", "Number of Trees (final)", "Tried score sets" ,"Best score sets" ))

# append them to the list
ordinalForest$parameters <- prm
```
Now that the initial code is written, I can define the default training grid. Some models can do a random search, but I wont implement that as the [paper] (https://epub.ub.uni-muenchen.de/41183/1/TR.pdf) states the defaults are pretty good.
```{R}
ordinalForestGrid <- function(x, y, len = NULL, search = grid) {
  if(search == grid) {
    out <- expand.grid(nsets = 1000,
                       ntreeperdiv = 100,
                       ntreefinal = 5000,
                       npermtrial = 500,
                       nbest = 10)
  } else {
    stop('random search not yet implemented')
  }
  out
}

# append to list
ordinalForest$grid <- ordinalForestGrid
``` 
For the fitting function, the `ordfit` function does have a strange implementation in defining `X` and `Y`. I try and overcome this by explicitly asking for `X` and `Y`, binding them and then implementing the `ordfit` function:
```{R}
ordinalForestFit <- function(x, y) {
  library(ordinalForest)
  
  data <- data.frame(x, Class = y)
  
  ordfor(depvar = "Class", data, nsets = 1000, ntreeperdiv = 100, ntreefinal = 5000,
         perffunction = c("equal"), classimp,
         classweights, nbest = 10, naive = FALSE, num.threads = NULL,
         npermtrial = 500, permperdefault = FALSE)
  
}

# append to list
ordinalForest$fit <- ordinalForestFit
```
[...]

Issue #890, opened 2018-05-24
qrf predict func 


olangfor:
Hi,

Think I may have spotted a bug in the quantile random forest regression method.  The predict function in the code passes in quantiles  = 0.5 as an argument.  I believe it is meant to be what = 0.5 (see ?quantregForest::predict.quantregForest).  A result of passing this in, the predict function is returning the quantile at 0.1.

Please let me know if I've missed interpreted this. 

Kind Regards,

Oliver

hadjipantelis:
Good catch! :+1: 

I checked this quickly and I think you have not misinterpreted something. 
I have a PR about it, it will be committed in the following days.

PR #921, opened 2018-07-26
Ordinal Forest and some minor issues

hadjipantelis:
This regarding issues #890 and #902. Some minor issues as mentioned in #902 might require fine-tuning.


topepo:
Thanks for catching this and the test errors.

zzw0034:
Hi, how could i use predict for qrf for quantile 0.1,0.9?
thank you.

Issue #930, opened 2018-08-21
findLinearCombos() throws an error when only constant variables present


michaelwhammer:
Minimal, reproducible example via `reprex::reprex()`:
```
library(caret)
#> Warning: package 'caret' was built under R version 3.4.4
#> Loading required package: lattice
#> Loading required package: ggplot2
#> Warning: package 'ggplot2' was built under R version 3.4.4
test_mat <- matrix(NA, nrow = 6, ncol = 4)

test_mat[, 1] <- 0
test_mat[, 2] <- 1
test_mat[, 3] <- rep(c(1, 0), each = 3)
test_mat[, 4] <- rep(c(0, 1), each = 3)
test_mat
#>      [,1] [,2] [,3] [,4]
#> [1,]    0    1    1    0
#> [2,]    0    1    1    0
#> [3,]    0    1    1    0
#> [4,]    0    1    0    1
#> [5,]    0    1    0    1
#> [6,]    0    1    0    1
## Works
findLinearCombos(test_mat)
#> $linearCombos
#> $linearCombos[[1]]
#> [1] 1
#> 
#> $linearCombos[[2]]
#> [1] 4 2 3
#> 
#> 
#> $remove
#> [1] 1 4

## Throws an error
findLinearCombos(test_mat[, 1:2])
#> Error in enumLC.default(x[, -badList]): enumLC does not support  numeric objects
```

Session Info:
```
sessionInfo()
#> R version 3.4.2 (2017-09-28)
#> Platform: x86_64-w64-mingw32/x64 (64-bit)
#> Running under: Windows 10 x64 (build 17134)
#> 
#> Matrix products: default
#> 
#> locale:
#> [1] LC_COLLATE=English_United States.1252 
#> [2] LC_CTYPE=English_United States.1252   
#> [3] LC_MONETARY=English_United States.1252
#> [4] LC_NUMERIC=C                          
#> [5] LC_TIME=English_United States.1252    
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> other attached packages:
#> [1] caret_6.0-80    ggplot2_3.0.0   lattice_0.20-35
#> 
#> loaded via a namespace (and not attached):
#>  [1] Rcpp_0.12.17       lubridate_1.7.4    tidyr_0.8.1       
#>  [4] class_7.3-14       assertthat_0.2.0   rprojroot_1.3-2   
[...]

PR #931, opened 2018-08-22
Minor fix for 930

hadjipantelis:
Minor fix for #930.


topepo:
Well that's embarrassing. If I could go back in time, I wouldn't go after Hitler, I would try to convince John Chambers to default this to `FALSE`. 

Thanks

PR #938, opened 2018-09-07
More statistics for function thresholder

BenjaK:
The function `thresholder` currently computes Sensitivity, Specificity, Youden's J and the distance to the best possible cutoff. However, these may not be the metrics that the user wishes to compare models by. 

Internally, `thresholder` uses `caret::confusionMatrix` which calculates a bunch of other statistics, but throws away all but those above. This pull request enables the user to choose which statistics to compute for each probability cutoff, the options being the same as those calculated in `caret::confusionMatrix`. The default option calculates all statistics from `caret::confusionMatrix`.


topepo:
Thanks

PR #949, opened 2018-10-10
Add support for non-atomic hyper parameters

wligtenberg:
Prophet has two hyper parameters that can be data.frames. This breaks certain things in caret.
This PR adds changes to caret to temporarely replace all non-atomic columns by numbers to allow sorting to work on the various different places they occur.

Basically I create a unique list of all items in a column and replace it by the index of that list.
We can then sort normally, and afterwards I replace them back.

I included a test with an example prophet wrapper, which I left in the test, because I was not sure if you want to have it in. Feel free to make it an official thing if you want. Otherwise the test can help people with their own implementation and use.

Feedback is welcome.


topepo:
I'm back to `caret` working on a new release and went back over this

>  I understand that you don want to make invasive changes to caret.

I'm not willing to take on more significant dependencies. There are a lot of model-specific changes here and I don't think that I've ever made this much for one specific model. 

I'm unwilling to accept the PR for this reason and because I inherit the responsibility of maintaining these changes (and their consequence). Since I don't do a lot with `prophet`, that is not something that I want to do. 

> Knowing that caret is somewhat in maintenance mode, I am very interested to see how we can make this happen in the newer packages.

For `parsnip` and `dials`, we are figuring out how to represent these types of tuning parameters. Maybe file an issue in `dials` with more specificity about the parameters that you want to represent and their structure.

PR #966, opened 2018-11-14
Fix bug in zv & corr preProcess

floriandeboissieu:
In preProcess with methods "zv" and "corr", if there is zero variance variable to remove, it must be taken into account for correlation computation, otherwise correlation fails to compute.


topepo:
Thanks. 

I'm going to add the additional step of adding the "zv" method when anyone chooses "corr" in the `method` argument.

PR #972, opened 2018-11-22
fix bug: na.remove in preProcess 'zv' method

floriandeboissieu:
na.remove was not taken into account in preProcess 'zv' method, leading to non-zv variables with only one value and NAs.

### DIFF
--- a/pkg/caret/R/preProcess.R
+++ b/pkg/caret/R/preProcess.R
@@ -236,9 +236,11 @@ preProcess.default <- function(x, method = c("center", "scale"),
   ## check for zero-variance predictors
   if(any(names(method) == "zv")){
     if(is.data.frame(x)) {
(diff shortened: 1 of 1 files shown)


pkg/caret/R/preProcess.R
```diff
@@ -236,9 +236,11 @@ preProcess.default <- function(x, method = c("center", "scale"),
   ## check for zero-variance predictors
   if(any(names(method) == "zv")){
     if(is.data.frame(x)) {
-      is_zv <- unlist(lapply(x[, !(colnames(x) %in% method$ignore), drop = FALSE], function(x) length(unique(x)) <= 1))
```
topepo: Please represent the true brach here as a logical and not integer (e.g. `length(na.omit(unique(x)))`)

topepo:
Can you please add a unit test for this too?

floriandeboissieu:
I replace

length(unique(x)) <= 1

by

ifelse(na.remove, length(na.omit(unique(x))), length(unique(x))) <= 1

so it is the same type as the original: logical.

Please make your request more accurate.

PR #973, opened 2018-11-22
fix bug: varImp for pls model with ncomp=1

floriandeboissieu:
when ncomp=1 in pls model weight matrix became a vector at weight.mean computation.
This was corrected with an if in the pls.R model file.


topepo:
Can you please add a unit test for this and update the NEWS file too?

topepo:
Thanks!

PR #985, opened 2019-01-13
Update mxnet.r

dimbov:
I think that replace should be changed to FALSE. 

> sample(c('relu', 'sigmoid', 'tanh', 'softrelu'), replace= TRUE)
[1] "**relu**"     "sigmoid"  "**relu**"     "softrelu"

> sample(c('relu', 'sigmoid', 'tanh', 'softrelu'), replace= FALSE)
[1] "softrelu" "relu"     "tanh"     "sigmoid"

### DIFF
--- a/models/files/mxnet.R
+++ b/models/files/mxnet.R
@@ -19,13 +19,13 @@ modelInfo <- list(label = "Neural Network",
                                          dropout = seq(0, .7, length = len), 
                                          activation = 'relu')
                     } else {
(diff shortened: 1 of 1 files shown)


topepo:
I don't think so. For random search, you may ask for > 4 tuning parameter combinations and this would reslut in 

```r
> sample(c('relu', 'sigmoid', 'tanh', 'softrelu'), replace= FALSE, size=5)
Error in sample.int(length(x), size, replace, prob) : 
  cannot take a sample larger than the population when 'replace = FALSE'
```

PR #987, opened 2019-01-15
Use :: in chaid_control

feebarscevicius:
This avoids getting a 'chaid_control not found' error when training CHAID without the 'control' argument.


topepo:
Thanks

PR #988, opened 2019-01-16
Typo on 'Notes on Reproducibility'

aneisse:
Where was written 'and this may effect reproducibility' I think the author wanted to write 'and this may affect reproducibility'.


topepo:
Thanks

PR #991, opened 2019-01-19
add space below kappa in print.confusionMatrix

malcolmbarrett:
As noted in https://github.com/tidymodels/broom/issues/523, the McNamara's test p-value is unrelated to kappa. I've added a space in `print.confusionMatrix` to separate the two for clarity, given that they are separate from accuracy


malcolmbarrett:
(travis fail on old release looks unrelated, unsurprisingly)

topepo:
> (travis fail on old release looks unrelated, unsurprisingly)

`earth` has a recent R dependency that makes this fail. 

Thanks

PR #1010, opened 2019-02-19
Prevent duplicates in discrete tune grids

dashaub:
Prevent duplicate tuning parameter combinations in discrete random grids. There are probably other models that need this too, but I only looked at the models that I frequently use.


topepo:
Instead, can you make this change to `train.default` right after the grid is created? I'd add it before the line 

```r
## Check to make sure that there are tuning parameters in some cases
```

and add `out[!duplicated(out), ,drop = FALSE]`. 

Also, please update the NEWS file.

dashaub:
@topepo will do. I believe we can then also remove the checks for duplicates `out <- out[!duplicated(out),]` in all the other [model files](https://github.com/topepo/caret/search?q=duplicated%28out%29&unscoped_q=duplicated%28out%29) too, right?

topepo:
> I believe we can then also remove the checks for duplicates `out <- out[!duplicated(out),]` in all the other [model files](https://github.com/topepo/caret/search?q=duplicated%28out%29&unscoped_q=duplicated%28out%29) too, right?

Yep. 

Thanks

dashaub:
Done. I think I got them all:
```
$ grep 'duplicated(out)' files/*.R
$ grep 'duplicated(out)' deprecated/*.R
```

dashaub:
@topepo Anything else you'd like on this?

topepo:
Thanks

Issue #390, opened 2016-03-10
caret::dummyVars reoccurring pattern in column name causes errors in dummy variable names


JanLauGe:
I noticed that dummyVars is producing erroneous variable names when creating (predicting) dummy variables if one of the column names in the original dataset matches the start of the name string of a subsequent column name. For these cases, the new dummy variable names get split in the wrong place. Part of the column names of the partly matching subsequent column name is put with the factor level name.
As far as I can tell the function still delivers the right result, just with a confusing name.

#### Minimal dataset:
```{R}
data <- data.frame('id' = seq(1,30,1),
                   'fooFactor' = factor(c(rep(1,10), rep(2,10), rep(3,10))),
                   'fooFactorBar' = factor(c(rep(4,10), rep(5,10), rep(6,10))),
                   'fooBarFactor' = factor(c(rep(7,10), rep(8,10), rep(9,10))))
```
#### Minimal, runnable code:
```{R}
library(caret)
library(dplyr)
#make some data
data <- data.frame('id' = seq(1,30,1),
                   'fooFactor' = factor(c(rep(1,10), rep(2,10), rep(3,10))),
                   'fooFactorBar' = factor(c(rep(4,10), rep(5,10), rep(6,10))),
                   'fooBarFactor' = factor(c(rep(7,10), rep(8,10), rep(9,10))))
#dummify the data
dummies <- dummyVars(formula = id ~., 
                     data = data,
                     sep = '-') %>%
  predict(data)

#check the names
colnames(dummies)
#will return:
# [1] "fooFactor-1"    "fooFactor-2"    "fooFactor-3"    "fooFactor-Bar4" "fooFactor-Bar5"
# [6] "fooFactor-Bar6" "fooBarFactor-7" "fooBarFactor-8" "fooBarFactor-9"

#notice how 'fooFactor' and 'fooBarFactor' are both fine,
#but 'fooFactorBar' gets turned into 'fooFactor-Bar4' etc.
```
The same is true when using 'levelsOnly = TRUE' by the way. With this option, dummy variable names become 1, 2, 3, Bar4, Bar5, Bar6, 7, 8, 9.

This is my first bug report on github. Please point out anything that is missing or should be done better. Thanks for all the effort that went into this fantastic and super helpful package!

### Session Info:
R version 3.2.3 (2015-12-10)
Platform: x86_64-apple-darwin14.0.0 (64-bit)
Running under: OS X 10.10.5 (Yosemite)

locale:
[...]

PR #1011, opened 2019-02-26
A quick fix of issue #390 partially solved in commit b4f8a87ce516b6f5d9d8a11cfb940a3b696ec5ca (see my recent comments on #390)

adimajo:
See issue #390 for a MWE.

dummyVars behaves badly when confronted with `colnames` "X1", "X10" which have reocurring patterns in their factor levels, e.g. "1", ... "10" for which `colnames` of the one-hot encoded resulting matrix are "X1.1.1".

This fix adds `stringi` to the Imports list, which is probably overkill, and uses a `which` statement which might result in a performance bottleneck.

### DIFF
--- a/pkg/caret/DESCRIPTION
+++ b/pkg/caret/DESCRIPTION
@@ -15,6 +15,7 @@ Depends:
 URL: https://github.com/topepo/caret/
 BugReports: https://github.com/topepo/caret/issues
 Imports:
+    stringi,
     foreach,
     methods,
     plyr,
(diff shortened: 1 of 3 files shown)


topepo:
Can you use base R functions instead of `stringi`? That is "like the heaviest of heavy dependencies"

adimajo:
I got rid of it in my last commit and added a test in the appropriate
testthat file.

Le lun. 25 mars 2019 à 2:39 AM, Max Kuhn <notifications@github.com> a
écrit :

topepo:
Thanks!

PR #1038, opened 2019-05-22
fixed bug that was causing duplication in data

adosib:
See https://stackoverflow.com/questions/56262318/caretcalibration-duplicating-data-when-including-multiple-model-columns-in-the.


topepo:
Thanks

PR #1054, opened 2019-07-16
Fixed documentation of list argument in downSample

Alaburda:
Originally: "list - should the function return list(x, y) or bind x and y together? If TRUE, the output will be coerced to a data frame." I think it should say FALSE instead.


topepo:
Great! Thanks

Issue #872, opened 2018-04-18
Error in varImp[, "%IncMSE"] on 'rf' model


floriandeboissieu:
varImp sends error if trained on  "rf" without specifying importance=TRUE. As importance may be an argument for other models, why not adding it explicitly to rf.R fit function, line 19:
```
fit = function(x, y, wts, param, lev, last, classProbs, ...)
                    randomForest::randomForest(x, y, mtry = param$mtry, importance = TRUE, ...)
```

Here is an example:
```
library(caret)
library(randomForest)

# Simulate data
dat <- caret::SLC14_1(1e2)
x <- dat[,c(1:5)]
y <- dat$y
ctrl <- trainControl(method = 'cv')
# varImp sends error 
fit_cont <- train(x, y)
varImp(fit_cont)

# importance=T works
fit_cont <- train(x, y, importance=T)
varImp(fit_cont)
```

topepo:
While that does deem sensible, it's not defaulted to `TRUE` for the same reason that the underlying model doesn't automatically compute it: it can take forever to run. 

I've done the same thing but OTOH I don't usually compute that until I know that I want to use that model. I'm going to leave it the way that it is. You can always use a custom method to with the code that you have above to get what you want.

PR #1063, opened 2019-08-10
Fix `varImp()` for `rf` regression model

sergeykorop:
This is an alternative approach to issue #872: when `rf` regression model is trained with `importance = FALSE` and `%IncMSE` column is missing in the output we can use an alternative importance score similarly to classification case.

I didn't update `pkg/caret/inst/models/models.RData` intentionally since this binary file can't be properly reviewed and, therefore, IMHO it should be only updated by repository owner or other trusted committer.


topepo:
Thanks

Issue #474, opened 2016-08-21
Sparse Matrix in caret


farbodr:
It is possible to use sparse matrix with caret? All the code that I've seen in caret uses as.matrix(x) which converts it to a dense matrix (as it was pointed out to me on stack overflow).  This sort of defeats the purpose of using a sparse matrix. 

TIA,
FR

topepo:
> All the code that I've seen in caret uses as.matrix(x) which converts it to a dense matrix 

`'train.default` doesn't do that. There are likely models that require a non-sparse representation of the data but I'm pretty sure that `train` leaves it alone. Most/all of any required conversion code happens in the model code and only when the model requires (AFAIK). 

Some of that model code was written a decade ago; if a model now allows for a sparse matrix, let me know or do a PR. 

Thanks,

Max

farbodr:
Thanks Max. One that comes to mind is Ranger. I think it can take a sparse matrix for x but from what I can tell from fit method in caret it gets converted to data frame.  I 'm not quite sure but I think nnet can also accept a matrix but it gets converted to data frame in caret.

XGBoost and glmnet are ones that I'm using right now and both support sparse matrix.
 
FR

farbodr:
I take that back I don't think ranger supports matrix.

topepo:
The `xgBoost` methods already converts the data to a sparse format using `xgb.DMatrix`. I added code to avoid re-converting. 

In `glmnet`, I also avoided the conversion too. 

Test away!

farbodr:
Thank you for speedy response. I noticed line 56 in glmnet.R may not be correct. It says 'x <- as.matri(x)' and should be x <- as. matrix(x).

FR

topepo:
Thanks

farbodr:
I think one more change is needed:

```
if(!(class(x)[1] %in% c("matrix", "sparseMatrix")))
                      x <- as.matrix(x)
```
should be
```
if(!(class(x)[1] %in% c("Matrix", "sparseMatrix")))
                      x <- as.matrix(x)
```

topepo:
I'm not sure. `?glmnet` has for `x`:

 > input matrix, of dimension nobs x nvars; each row is an observation vector. Can be in sparse matrix format (inherit from class "sparseMatrix" as in package Matrix; not yet available for family="cox")

It needs to be a matrix of some sort and I interpret the above to mean either "matrix" or "sparseMatrix"

schistyakov:
[...]

PR #1096, opened 2019-11-18
Train a glmnet model with a sparseMatrix

maelick:
This fixes caret's glmnet model object so that sparse matrix objects are not converted to regular matrix anymore. I was experiencing extremely long CV and training times with caret in comparison to vanilla `glmnet::cv.glmnet` with sparse matrix data coming from text2vec.

This problem was mentioned in #474. Caret's code for training and testing a glmnet model currently checks if the input data is a sparse matrix using the `class` function. This is incorrect as sparse matrices from Matrix can be of different classes. For example, my object returned by text2vec is of class `dgCMatrix`. The fix relying on the `inherits` function instead of `class` should work with any sparse matrix returned by the Matrix package.

(Note that I didn't include in the commit the updated binary pkg/caret/inst/models/models.RData generated by models/parseModels.R)

### DIFF
--- a/models/files/glmnet.R
+++ b/models/files/glmnet.R
@@ -52,20 +52,19 @@ modelInfo <- list(label = "glmnet",
                     ## pass in any model weights
                     if(!is.null(wts)) theDots$weights <- wts
 
-                    if(!(class(x)[1] %in% c("matrix", "sparseMatrix")))
+                    if(!is.matrix(x) && !inherits(x, "sparseMatrix"))
                       x <- Matrix::as.matrix(x)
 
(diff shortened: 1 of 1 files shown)


topepo:
Can you add some unit tests for this and #1097?

maelick:
How should I test for this one?
I manually checked it worked using my own data with thousands of rows and columns and for which there was a huge speed improvement after the change. However, the original data returned by train.default is converted to a data.frame with the fix from #1097 and is not returned at all without. One solution is to not convert Matrix objects to data.frame in train.default. Is there any reason justifying this conversion or would it be fine to keep it as a Matrix?

PR #1097, opened 2019-11-18
Conversion of sparse matrix data to data.frame in train.default when returnData is TRUE

maelick:
When `returnData` is `TRUE` in `trainControl`, and the training data is a sparse matrix object, the following warning is generated:

`Warning train.default(x = train_x, y = as.factor(ifelse(train_y == 1,  :
    The training data could not be converted to a data frame for saving`

This fixes the problem (mentioned in #474) by adding a simple condition checking for `sparseMatrix` objects in `train.default` and converts it to regular `matrix` object before conversion to `data.frame`.


maelick:
I added a test checking that train doesn't raise warnings using matrices from base R and from the Matrix package.

PR #1099, opened 2019-11-20
Update twoClassSummary

mlduarte:
I believe specificity is looking at the incorrect level?


topepo:
No. I intended that the first levels should be used.

Issue #1102, opened 2019-11-23
Speed up groupKFold 


mayer79:
Hello. The function `groupKFold` is very important in practical applications. However, it is extremely slow. 

Demo code:

```{R}
X <- data.frame(group = sample(1:1e6, size = 3e6, replace = TRUE))
system.time(out <- caret::groupKFold(X$group, k = 5)) # 53 Seconds
```

Here is a fast version (with slightly different output structure):
```{R}
fastGroupKFold <- function(group, k) {
  g_unique <- unique(group)
  stopifnot(k <= length(g_unique))
  g_folds <- sample(k, size = length(g_unique), replace = TRUE)
  split(seq_along(group), g_folds[match(group, g_unique)])
}
system.time(out <- fastGroupKFold(X$group, k = 5)) # 0.6 Seconds
```

It would be fantastic if you could speed up the current implementation of `groupKFold`.

topepo:
If you would make the output format consistent with what we would need, please feel free to make a PR.

topepo:
closed via #1108

PR #1108, opened 2020-01-02
accelerated groupKFold

mayer79:
Implements https://github.com/topepo/caret/issues/1102#issue-527542475

In- and output interface remain unchanged, but the results are not backward compatible as the splits are done differently.


topepo:
👍 Thanks

PR #1120, opened 2020-02-25
remove space before a full stop

paulkaefer:
Minor, I know, but it was bugging me as I was reading the code while trying to solve a modeling issue.


topepo:
I'm going on vacation for a few weeks. I'll take a look when I return. Thanks!

PR #1162, opened 2020-07-15
Update createModel.R

kransom14:
Return model data and response variable in model object. This way both will be available in final model object created by train. Allows for more transparency when sampling routines or preprocessing is performed on the data before modeling.

### DIFF
--- a/pkg/caret/R/createModel.R
+++ b/pkg/caret/R/createModel.R
@@ -59,6 +59,8 @@
     modelFit$tuneValue <- tuneValue
     modelFit$obsLevels <- obsLevels
     modelFit$param <- list(...)
+    modelFit$data <- x
+    modelFit$response <- y
   }
 
   list(fit = modelFit, preProc = ppObj)


topepo:
`train()` doesn't save the individual models, so this wouldn't affect anything but the final fit object. For all of the other invocations, it will consume memory for no real reason. If you want it for the `finalModel` object, you could wrap it inside a conditional statement using `last`. 

However, we do save the processed version of the data in `object$trainingData` and the original data could be attached after the fact. TBH I don't see the utility in this change.

kransom14:
Maybe createModel isn't the place to save the processed data as I see what you mean about the memory consumption as createModel is run within CV tuning. However, the processed data is not saved anywhere that I can see when subsampling routines are used (but maybe only for other preprocessing steps?). For example, when I check object$trainingData on the following example it does not have the subsampling applied to it when I use the options for sampling (for example "down" for down sampling):

```
library(caret)

# make an example imbalanced data set (from https://topepo.github.io/caret/subsampling-for-class-imbalances.html#subsampling-techniques)
set.seed(2969)
imbal_train <- twoClassSim(10000, intercept = -20, linearVars = 20)
imbal_test  <- twoClassSim(10000, intercept = -20, linearVars = 20)
table(imbal_train$Class)

# make a tuning grid
xgbGrid <- expand.grid(nrounds = c(500, 1000), 
                       max_depth = c(9),
                       eta = c(0.01),
                       gamma = c(0,1),
                       colsample_bytree = c(0.5),
                       min_child_weight = c(9),
                       subsample = c(0.5))

fitControl_down <- trainControl(method="repeatedcv",
                                number = 10,repeats = 1,
                                returnData = TRUE,
                                verboseIter = FALSE,
                                classProbs = TRUE,
                                allowParallel = TRUE,
                                sampling = "down")

# down sampling within cv tuning
set.seed(825)
xgbTune_down <- train(Class~., data = imbal_train,
                      method = "xgbTree",
                      metric = "Accuracy",
                      tuneGrid = xgbGrid,
                      trControl = fitControl_down,
                      verbose = FALSE,
                      na.action = na.pass) 

c <- xgbTune_down$trainingData

summary(c$.outcome) # still imbalanced
identical(c[,2:26], imbal_train[,1:25]) # identical to the original data
```

topepo:
Sorry but I'm not willing to include this as a feature. It may be a pain, but you could do this on an as-needed basis by copying the model definition code (produced by `getModelInfo()`) and modifying it to save the data (say to a temp directory). I see this as something that would be good for debugging but is otherwise out-of-scope for `caret`.

kransom14:
No problem, it does seem more efficient to get the data out on an as needed basis if it causes slow-downs.

kransom14:
No problem, it does seem more efficient to get the data out on an as needed basis if it causes slow-downs.

PR #1179, opened 2020-10-29
Fix "var_seq(..., len=1)" - regression/classification defaults were swapped

hendrikweisser:
From the `var_seq` documentation:
> If `len = 1`, the defaults from the `randomForest` package are used.

From the `randomForest` documentation:
> Note that the default values are different for classification (sqrt(p) where p is number of variables in 'x') and regression (p/3)

The right values were used, but for the wrong case.


topepo:
Thanks!

PR #1186, opened 2021-01-02
Corrected the definition of R^2 in the man page.

taroyabuki:


topepo:
Thanks! In the future, keep in mind that the Rd files are automatically generated from the R files. It's better to edit the R file to make the change permanent.

PR #1188, opened 2021-01-11
Change default objective to 'reg:squarederror' for xgboost models 

rajkar86:
To supress reg:linear deprecation warning.

The other option is to just exclude this parameter, since since it's the default anyway for xgboost.train.
 
That would have the benefit of not printing "parameters were provided multiple times" warning should the user try to use a different objective, but it makes the caret usage inconsistent between regression and classification.


topepo:
Looks great. Thanks!

PR #1210, opened 2021-03-10
Correction in mlpKerasDecay for regression

DominiqueCHAPELON:
In line 81, the compile function is called without the prefix keras::
If you try to build a regression model, the following error is returned :
compile function not found

Note :
I tried on my local with the correction and it works now.


topepo:
Excellent. Thanks!

PR #1225, opened 2021-05-11
GHA update for 2021-05 and rmarkdown to suggests

topepo:


topepo:
Going to merge this despite the failures. 

It initially failed on Windows/R 3.6 since it didn't have `klaR` installed (despite being in Suggests). It seems like a random issue and the new GHA use `pak`to install (which should be more robust) 🤞

Issue #1226, opened 2021-05-12
use themis::smote


topepo:
The `DMwR` package is archived due to non-responsiveness. We should move to `themis::smote()` instead.

EmilHvitfeldt:
I don't know if it will throw a wrench in this, but `smote()` is not exported from {themis} right now.

If you need I can export and do a CRAN update fairly fast

topepo:
Ok. It would be good to export that so we can use it. I can do another release soon once that is changed and I've made the corresponding `caret` changes.

hermandr:
Hi, when will caret be updated with themis::smote()?

EmilHvitfeldt:
{themis} have been updated and pushed to CRAN
https://github.com/tidymodels/themis/releases/tag/v0.1.4
https://cran.r-project.org/web/packages/themis/index.html

PR #1252, opened 2021-09-20
Sub-Sampling changes

topepo:
Two changes: 

 * `themis::smote()` is now used since `DMwR` is archived/orphaned. 
 * In the new version of `recipes`, `step_upsample()` has been deprecated and removed.


topepo:
closes #1226

PR #1268, opened 2021-12-16
Harmonize ranger rf hyperparams

jttoivon:
Harmonize the hyperparameters of ranger and rf a bit: added 'nodesize' as hyperparameter for rf. Added hyperparameter num.trees/ntree to ranger/rf like in tidymodels.

When I tested this on my data, it worked, but ranger's performance got better with larger min.node.size, whereas rf's performance got worse with larger nodesize. Maybe this was due to the hellinger splitrule used for ranger?


topepo:
Thanks for the contribution and sorry for the long wait. 

I feel pretty strongly that the number of trees is not a tuning parameter and should be set to some high number. That said, I've included it as a value that can be changed but defaulted regular and non-regular grids to a single value (500 trees). That means that people _can_ tune it but caret won't directly facilitate that.

topepo:
@jttoivon Sorry, this change broke a bunch of dependencies. 

I have to submit caret today so I parked the changes in a branch and will redo them on GH after the release. From there, I'll contact the other package maintainers and have them make the changes required for the extra RF parameters.

PR #1274, opened 2022-03-02
Meaningful error message when using prSummary with classProbs=FALSE

xabriel:
When `classProbs=FALSE`, `prSummary()` fails without letting the user know what to do.

This PR makes the `prSummary()` function fail similarly as `twoClassSummary()` when a user forgot to set `classProbs=TRUE`.


xabriel:
@topepo just wanted to ping you in case you missed this PR. Let me know if it is helpful. Thanks.

topepo:
Thanks for the contribution and waiting so long. I'll update from main and merge 👍

Issue #1279, opened 2022-04-10
For XGBOOST can caret utilize "gpu_hist" for xgboost compiled with GPU support?


aadler:
If I have finally successfully built `xgboost` with GPU support on Windows, so "gpu_hist" is valid when I run it directly in the `xgboost` package, is there any way to have caret access this functionality in the `xgbTree` method? It's near 30X faster on my computer! If not, can this become a feature request?

aadler:
@topepo I haven't yet figured out how to build *caret* on my machine given its structure, but I'm hoping this simple PR is sufficient: #1310. If you could point me to how to build the package locally I can test personally (I built XGBOOST with GPU support, see https://github.com/dmlc/xgboost/issues/8190) but if the CI works that would indicate it's good, no?

leedrake5:
It's not a caret problem, it's an XGBoost thing. You need to install a version of it with GPU support. They have os-specific binaries that will do it for you: 

> "We also provide experimental pre-built binary with GPU support. With this binary, you will be able to use the GPU algorithm without building XGBoost from the source. Download the binary package from the Releases page. The file name will be of the form xgboost_r_gpu_[os]_[version].tar.gz, where [os] is either linux or win64. (We build the binaries for 64-bit Linux and Windows.) Then install XGBoost by running:

> '''xgboost, Release 2.0.0-dev
    # Install dependencies
R -q -e 
> install.packages(c('data.table', 'jsonlite'))" 
> # Install XGBoost
> R CMD INSTALL ./xgboost_r_gpu_linux.tar.gz"'''

Releases page here: https://github.com/dmlc/xgboost/releases

Information in what model configurations have gpu support here: https://xgboost.readthedocs.io/en/stable/gpu/index.html

Note that caret is just a wrapper - I forked my own version to add max_delta_step and scale_pos_weight as variables: https://GitHub.com/leedrake5/caret. 

Have to stress - caret is just a nice wrapper. Everything needs to be done in XGBoost. 

Once everything is properly installed, make sure to use tree_method="gpu_hist" as an argument in train(). Any additional arguments you pass will by default go to xgboost.train, you can find those arguments here: https://xgboost.readthedocs.io/en/stable/parameter.html

aadler:
@leedrake5 I understand. I'm sorry if I wasn't clear but I thought the l;inks I provided would explain.

[...]

PR #1310, opened 2022-09-05
Adding 'tree_method' to access 'hist' and 'gpu_hist'

aadler:
I'm thinking this should allow people to use 'hist' and 'gpu_hist' given https://xgboost.readthedocs.io/en/stable/parameter.html#parameters-for-tree-booster. Package doesn't build on my machine probably because it is in non-standard format. Would you kindly test that this should cause no change?


aadler:
Unecessary; see [#1279](https://github.com/topepo/caret/issues/1279).

PR #1327, opened 2023-02-21
Update svmRadial.R

ianchi:
When computing class probabilities fail, the returned NA reusult have n times the number of rows. Categories are already defined as ncol, so they should not be repeated as rows


topepo:
@ianchi thanks for finding this

gmonaie:
I think this issue exists within svmLinear as well, the exact same $prob method exists there

PR #1361, opened 2024-04-22
Use full argument name for situations with restricted partial matching

MichaelChirico:
```r
options(warnPartialMatchArgs=TRUE, warn=2L)
train(mpg ~ cyl, data=mtcars)
# Error in stats::model.frame(form = mpg ~ cyl, data = mtcars, na.action = na.fail) : 
#   (converted from warning) partial argument match of 'form' to 'formula'
```

Similar for `seq(along=)` and `seq(length=)`.


MichaelChirico:
BTW, personally I just use `seq_along()` instead of `seq(along=)` and `seq_len()` instead of `seq(length=)`.

I can make the change to use those instead if you prefer.

topepo:
I have the same preference but

older_codes

We're good. Thanks

PR #1366, opened 2024-05-10
Update nnet.R - Correct the linout parameter

bpuladi:
Correct the linout parameter for nnet. This is required for regression. If not set, regression performance metrics will be wrong.


bpuladi:
Further changes

Issue #1419, opened 2026-06-09
Restructuring repository


topepo:
See #1418 

1. Move to topepo/caret-archive:
	- `RegressionTests`
	- `models/deprecated`
	- `html`
	- `deprecated`
	- `docs`

2. Move `pkg/caret/*` to top level

3. Move to `inst`, then add to `.Rbuildignore`
	- `bookdown` 
	- `models` (rename to `model_sources` to avoid existing `models` directory)
	- `release_process`
4. Update GHA files to stop pointing to the old location of the package

PR #1420, opened 2026-06-11
Move RegressionTests, models/deprecated, html, deprecated, docs to caret/archive

CharlotteZZh:
#1419

Archived to topepo/caret-archive with full history preserved (312 commits).


topepo (changes requested):
Looks pretty good. Here are two things that showed up when I ran `R CMD check`: 

1. There are empty directories that can be removed by us to avoid these notes: 

```
   Removed empty directory ‘caret/RegressionTests’
   Removed empty directory ‘caret/html’
   Removed empty directory ‘caret/pkg/caret/inst’
   Removed empty directory ‘caret/pkg/caret/tests’
   Removed empty directory ‘caret/pkg/caret’
   Removed empty directory ‘caret/pkg’
```

2. `.Rbuildignore` should have a few more entries. Take a look at `usethis::use_build_ignore()`. The entries are: 

```
N  checking top-level files ...
   Non-standard files/directories found at top level:
     ‘bookdown’ ‘models’ ‘release_process’
```

3. The GitHub actions failed because it's looking for the package in the old location: 

> working directory '/home/runner/work/caret/caret/pkg/caret'. No such file or directory

To fix this, go to `.github/workflows` and remove lines that have: 

> `working-directory: pkg/caret`

After the GHA will probably go smoothly. 


Otherwise, the _local_ `R CMD check` results were: 

> 0 errors ✔ | 0 warnings ✔

PR #1423, opened 2026-06-12
Move bookdown, models, release_process into inst/ (#1419)

CharlotteZZh:
- bookdown -> inst/bookdown
- models -> inst/model_sources (renamed to avoid the runtime inst/models)
- release_process -> inst/release_process
- Build-ignore the three under inst/ (dev material, kept out of the tarball)
- Fix path references in moved maintainer scripts and stale pkg/caret paths (parseModels.R, make_model_Rd.R, reverse_depends.R, READMEs, install hints)


topepo (approved):
Looks great. 11/10, no notes

PR #1441, opened 2026-06-18
Air format test files

CharlotteZZh:
## Summary
- Ran `air format` on `tests/` (air 0.9.0), reformatting 37 test files.
- Added `# fmt: skip` to large pre-computed expected-value vectors so air doesn't explode them into one-element-per-line:
  - `test_Dummies.R`: `exp_main_nomissing`, `exp_main_missing`, `exp_main_omit`
  - `test_preProcess_methods.R`: the two large `yj_pred_exp` structures
- Used leading `# fmt: skip` comments (trailing ones only suppress single-line statements in air).

## Notes
- Small 5-element `yj_lambda_exp` vectors were left as-is — minor, readable expansion rather than a "weird large" one.


topepo (approved):
Looks great. Merge if you are finished.

PR #1442, opened 2026-06-19
Testthat3

CharlotteZZh:
Run usethis::use_testthat(3)


topepo (approved):
Nice job! That went more smoothly than I thought it would.

Issue #1418, opened 2026-06-09
upkeep plan for summer 2026


topepo:
We have a paid intern this year who is interested in helping modernize caret to make support and upkeep easier. The current overall plan of no new features is still in place. We just want to make the repository saner and easier to work with. 

This issue is to help refine a short-term plan to start moving in this direction. It's a working document so please add your thoughts about what would improve the current repository without affecting reverse dependencies. 

# Initial plan (through the end of summer)

1. Restructure the repo [DONE]:
	- Move the actual package to the top level
	- Fix links (eg. GHA, model file location change)
	- Archive files that we probably don't need anymore, like the regression tests or the RHTML docs
	- Move a number of directories to `inst` and add them to `.Rbuildignore`

2. Update GitHub Actions [DONE]. Only 6mo behind, so this shouldn't be a big deal. Will probably add actions for java install etc [DONE]. 

3. PR triage: We don't want to create more work than required so I'll work through a fair number of PRs with a focus on PRs for unit tests [DONE]. 

4. `air format` [DONE]

5. Overhaul _existing_ tests [DONE]

	- Move to edition 3 of testthat
	- Helper files
	- No data/code outside of tests
	- Remove noise (e.g. logging, pkg load notifications) in tests
	- Snapshot tests
	- "NoSuggests" tests

6. Increase code coverage: We want to get the unit test coverage as good as possible, not only because it is a good idea, but also so that we can avoid Claude making mistakes later when we get help for refactoring [DONE]. 

# Code refactoring

After these steps, there is a laundry list of things to improve (in no particular order): 

- Move away from older technologies (plyr, etc)
- Code consolidation to reduce redundancy
- Refactoring and redocumenting
- _Perhaps_ moving to better logging of messages, warnings, and errors (to cli, if at all)
- Assess model list for archived or otherwise difficult dependencies. It'll be a high bar to remove them. 
- Convert bookdown docs to Quarto (mostly done) and maybe automate deployment. 

Again, the goal is to keep the interface the same and minimize reverse-dependency issues down the line. There will be some breakages because I'm continually surprised by the "off-label use" of dependencies. However, assuming we don't have to submit to "safely retain [our] package on CRAN", we should have time to submit PRs to other repos to deal with any changes.

DavisVaughan:
For air, you'll probably want to include a `.git-blame-ignore-revs` file https://github.com/posit-dev/air/issues/391

Example https://github.com/r-lib/vctrs/blob/main/.git-blame-ignore-revs

PR #1444, opened 2026-06-21
Tests: snapshot expectations + seed hygiene

CharlotteZZh:
Part of the test-suite overhaul (#1418, step 5). Builds on the edition-3 work.

## 1. Convert condition expectations to snapshots
- `expect_warning(x)` → `expect_snapshot_warning(x)`
- `expect_error(x[, "msg"])` → `expect_snapshot(x, error = TRUE)` (message regexes dropped — captured by the snapshot)
- **Edge case:** `expect_warning(x, NA)` / `expect_error(x, regexp = NA)` assert the *absence* of a condition, so snapshots don't apply — these became `expect_no_warning(x)` / `expect_no_error(x)` (in `test_sparse_Matrix.R`, `test_tibble.R`).

## 2. Seed hygiene (safe subset)
Replaced special-number seeds (`set.seed(1)`) with random integers in 1–10000, in tests whose assertions are structural / consistency / error-based: `test_bad_class_options`, `test_tibble`, `test_preProcess`, `test_Dummies`, `test_twoClassSummary`, `test_misc`. Each re-run to confirm FAIL:0 and no snapshot drift.

**Deliberately left unchanged** (would break hardcoded expected values tied to the seed; to be done when those assertions move to snapshots): `test_preProcess_methods`, `test_glmnet_varImp`, `test_BoxCox`, `test_multiclassSummary`, and snapshot-warning-dependent `test_safs`. The 8 `trim_*.R` files are excluded (testthat doesn't run them).

## Notes
- Incidental find: `test_confusionMatrix.R`'s `expect_error(confusionMatrix(dat5, ref1))` actually errored on a missing object (`ref1`), not the intended path — the snapshot now documents this; left as-is.


topepo (commented):
Looks great!

PR #1445, opened 2026-06-24
Tests: modern expectations + consistent skip_on_cran()

CharlotteZZh:
Part of the test-suite overhaul (#1418, step 5). Two related items:

## 1. Convert to modern expectations
- `expect_equivalent(a, b)` → `expect_equal(a, b, ignore_attr = TRUE)` (15×)
- `expect_is(x, "cls")` → `expect_s3_class(x, "cls")` (3×)
- `expect_that(x, is_a("cls"))` → `expect_s3_class()` / `expect_type()` (4×)
- `expect_true(inherits(x, "cls"))` → `expect_s3_class(x, "cls")` (4×)
- `expect_equal(length(x), n)` → `expect_length(x, n)` (6×)

## 2. Apply skip_on_cran() consistently
- Added `skip_on_cran()` as the first line of every `test_that()` block that lacked it — 49 blocks across 24 files.
- De-duplicated the ICA block (it already had one, after `skip_if_not_installed`).
- `test_BoxCox.R` had no `test_that` at all (ran via a top-level helper); wrapped its calls in a named `test_that()` so it's covered too.
- Result: all **82** test_that blocks now `skip_on_cran()`.

### DIFF
--- a/tests/testthat/test_BoxCox.R
+++ b/tests/testthat/test_BoxCox.R
@@ -29,8 +29,11 @@ check_BoxCox <- function(x, expected = NULL) {
   expect_equal(obs_lambdas1, expected)
 }
 
-check_BoxCox(dat, expected = exp_lambdas)
-check_BoxCox(
-  as.data.frame(dat, stringsAsFactors = TRUE),
-  expected = exp_lambdas
-)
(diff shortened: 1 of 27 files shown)


topepo (commented):
Looks great.

PR #1447, opened 2026-06-24
quiet tests and snapshots

CharlotteZZh:
1. Snapshot testing where output is static.
2. No stray printed output in devtools::test(),  wrapping library() in suppressPackageStartupMessages() removes the banner, leaving only the test results.

### DIFF
--- a/tests/testthat/test_BoxCox.R
+++ b/tests/testthat/test_BoxCox.R
@@ -2,25 +2,10 @@
 ## Generate data and do BC using the source function to get
 ## expected results
 
-suppressPackageStartupMessages(library(MASS))
-
 set.seed(1)
 dat <- matrix(runif(30), ncol = 3)
 dat[, 1] <- exp(dat[, 1])
 colnames(dat) <- paste0("x", 1:3)
-
-exp_lambdas <- rep(NA, 3)
-for (i in 1:ncol(dat)) {
-  tmp <- as.data.frame(dat, stringsAsFactors = TRUE)[, i, drop = FALSE]
-  names(tmp)[1] <- "x"
-  tmp_bc <- boxcox(
-    x ~ 1,
-    data = tmp,
-    plotit = FALSE,
-    lambda = seq(-2, 2, by = .1)
-  )
-  exp_lambdas[i] <- tmp_bc$x[which.max(tmp_bc$y)]
-}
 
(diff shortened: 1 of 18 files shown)


tests/testthat/test_bad_class_options.R
```diff
@@ -1,4 +1,4 @@
-library(caret)
+suppressPackageStartupMessages(library(caret))
```
topepo: Let's take out any `library(caret)` or `library(testthat)` calls in the test code. Those will happen automatically when the IDE's `test` buttons are used (RStudio or Positron)

tests/testthat/test_BoxCox.R
```diff
@@ -2,7 +2,7 @@
 ## Generate data and do BC using the source function to get
 ## expected results
 
-library(MASS)
+suppressPackageStartupMessages(library(MASS))
```
topepo: Two(ish) changes should happen in this case (where a package like MASS is used but is listed as a suggested package):

 - add a `skip_if_not_installed("MASS")` at the top of the test code block
 - For this example, MASS is indirectly loaded so we don't have to fully load it.
 - If it were used in the test code, we would want to namespace all the function calls (a la `MASS::lda()` etc) instead of fully loading it. 

Go ahead and look at the other instances and, if a package is in suggests, use the strategies above.

tests/testthat/test_tibble.R
```diff
@@ -1,7 +1,7 @@
-library(caret)
-library(recipes)
-library(dplyr)
-library(testthat)
+suppressPackageStartupMessages(library(caret))
+suppressPackageStartupMessages(library(recipes))
```
topepo: This instance is a little more complicated. 

recipes is an imported package, so we don't need a conditional skip as we did for MASS. We would end up using a number of functions from recipes (or ggplot2 or lattice) and, in this case, I would quietly load these imported packages in a helper file. So add a `helper-startup.R` file to the `testthat` directory and move these to that file. 

For dplyr, we would want a `skip_if_not_installed("dplyr")` here and do things like 

```
data = dplyr::as_tibble(df),
```

topepo (changes requested):
There's some nuance to which packages are loaded fully etc. I've made comments below but slack me or ask questions here if you have questions.

PR #1461, opened 2026-07-02
docs: fix links broken by the caret-archive move (#1420)

CharlotteZZh:
Repoint stale master URLs: relocated model files -> caret/main/inst/model_sources/files; archived content (html, deprecated, models/deprecated) -> topepo/caret-archive/main. The caret/main links resolve once the default-branch rename lands.


topepo (approved):
Good catch. I would have thought about this much later.

PR #1463, opened 2026-07-07
Consolidate imports into a usethis namespace block in R/caret-package.R

CharlotteZZh:
Gather the @importFrom / @import roxygen tags that were scattered across ~53 files into a single deduplicated block at the top of R/caret-package.R, wrapped in the standard usethis namespace markers:

    ## usethis namespace: start
    ...
    ## usethis namespace: end

Tags are grouped by package and wrapped to respect the 80-column line width. Verified with roxygen2 that the generated NAMESPACE is byte-identical, so this is a pure source refactor with no change to imported symbols.

### DIFF
--- a/R/aaa.R
+++ b/R/aaa.R
@@ -17,7 +17,6 @@
 
 
 #' @useDynLib caret
-#' @import methods plyr reshape2 ggplot2 lattice nlme
 NULL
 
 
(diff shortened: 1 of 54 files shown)


topepo (approved):
💯 Very nice and gratifying

PR #1467, opened 2026-07-10
Add tests for sensitivity, specificity, posPredValue and negPredValue

CharlotteZZh:
Cover the factor, table and matrix methods along with error paths, prevalence handling, NA handling and multiclass-table collapse for the sensitivity family of performance metrics. Ran air format.


tests/testthat/test-sensitivity.R
```diff
@@ -0,0 +1,142 @@
+# Reference data from the sensitivity() documentation example
+lvs <- c("normal", "abnormal")
```
topepo: Let's put these (and anything outside of the test blocks) in a helper

PR #1468, opened 2026-07-12
Fix recipe sbf boot632 resampling

sims1253:
Recipe-based SBF referenced an undefined data object and loop-local transformed values when calculating the apparent error for boot632 resampling. This caused the workflow to fail before returning results.\n\nReuse the existing seeded AllData resample as the apparent estimate and exclude it from the bootstrap average before applying the .632 weighting. The regression test now verifies the resulting weighted metrics.

Found with https://github.com/sims1253/ry

### DIFF
--- a/R/selectByFilter.R
+++ b/R/selectByFilter.R
@@ -642,9 +642,9 @@ sbf_rec <- function(rec, data, ctrl, lev, ...) {
   loadNamespace("caret")
 
   resampleIndex <- ctrl$index
-  if (ctrl$method %in% c("boot632")) {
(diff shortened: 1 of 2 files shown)


topepo (commented):
Looks good. Can you run `air format` on the code? [air](https://posit-dev.github.io/air/formatter.html) is the code formatter that we use

sims1253:
I hadn't done that as it completely reformats the entire file/project.

topepo:
You should update from main; we recently merged PRs that have the formatting changes

topepo (approved):
Thanks

PR #1469, opened 2026-07-13
Add tests for recall, precision, F_meas and prSummary

CharlotteZZh:
Adds `tests/testthat/test-prec_rec.R` (with fixtures in `tests/testthat/helper-prec_rec.R`) covering `R/prec_rec.R`, which previously had no unit tests.

Coverage includes:

- `recall`, `precision`, and `F_meas` for both the `table` and `default` (factor) methods
- Error paths (non-factor input, wrong number of levels)
- `F_meas` beta weighting
- `NA` handling (`na.rm`) across all three metrics, and the empty-relevant-denominator case returning `NA`
- Multiclass-table collapse
- `prSummary`: its AUC / precision / recall / F output and all three input-validation errors, guarded with `skip_if_not_installed("MLmetrics")`


tests/testthat/helper-prec_rec.R
```diff
@@ -0,0 +1,24 @@
+# Shared test data for test-prec_rec.R
```
topepo: Let's combine these and the sens/spec helpers in to a helpers-2x2-tables.R file

PR #1470, opened 2026-07-14
Add tests for maxDissim, minDiss and sumDiss

CharlotteZZh:


tests/testthat/helper-maxDissim.R
```diff
@@ -0,0 +1,11 @@
+# Shared test data for test-maxDissim.R
+
+# One "base" point at the origin plus a pool of points along the x-axis at
+# known distances (1, 5, 10), so the maximum-dissimilarity choices are
+# deterministic and can be worked out by hand.
+md_a <- matrix(c(0, 0), nrow = 1)
```
topepo: Since you say later on which are most dissimilar, maybe add some comments (or results, commented out) that show that this is the case for the example data. Otherwise, it's easy to overlook these and assume from the comments that it is true

topepo: Also, let's make more descriptive and unique names; we are adding a lot of helpers, and I don't want to have an inadvertent name overlap.

PR #1473, opened 2026-07-17
Fix model-diversity filter bugs in adaptive resampling and add tests

CharlotteZZh:
Three related bugs in the internal helpers used by adaptiveWorkflow():

- `cccmat()` used `if (i > j)` inside a `for (j in i:p)` loop, so the body never ran and it always returned a matrix of ones.
- `cccmat()` and diffmat() ran `out[lower.tri] <- out[upper.tri]` on an already-symmetric matrix; the two triangles index in different orders, which scrambled entries (and broke symmetry) for four or more models.
- `filter_on_diff()` and `filter_on_corr()` indexed the reordered column names with `newOrder[deletecol]`, permuting the selection twice and removing the wrong models.

Add test-adaptive.R / helper-adaptive.R covering `ccc`, `cccmat`, `diffmat`, `long2wide`, `get_id`, `filter_on_corr` and `filter_on_diff`, plus a NEWS entry.

### DIFF
--- a/tests/testthat/test-adaptive.R
+++ b/tests/testthat/test-adaptive.R
@@ -1,0 +1,137 @@
+# The fixtures (adapt_results, adapt_wide, adapt_grid) live in helper-adaptive.R.
+# These check the little model-picking helpers behind adaptiveWorkflow() - the
+# workflow itself is tested elsewhere.
+
+# --- ccc --------------------------------------------------------------------
+
+test_that("ccc returns Lin's concordance correlation coefficient", {
+  # two identical vectors agree perfectly, so we should get exactly 1
+  expect_equal(caret:::ccc(1:5, 1:5), 1)
+  # a pair that disagrees, with a value we can work out by hand
+  expect_equal(caret:::ccc(1:5, c(5, 3, 1, 2, 4)), -0.3)
+  # shifting every value by a constant should pull ccc below 1, even though the
+  # plain correlation would still be a perfect 1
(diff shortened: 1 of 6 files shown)


EmilHvitfeldt (approved):
Very nice, good clean PR

EmilHvitfeldt:
do we have any issues that we can link to for this PR or was it just random things you found while you were working here?

tests/testthat/test-adaptive.R
```diff
@@ -0,0 +1,137 @@
...
+test_that("filter_on_corr keeps everything when nothing exceeds the cutoff", {
+  keep <- sort(unique(
+    caret:::filter_on_corr(adapt_results, "RMSE", cutoff = 1)$model_id
+  ))
+  expect_equal(keep, c("m1", "m2", "m3", "m4"))
+})
+
+test_that("filter_on_corr errors with a single model", {
+  one <- adapt_results[adapt_results$model_id == "m1", ]
+  expect_error(caret:::filter_on_corr(one, "RMSE", cutoff = 0.9))
```
EmilHvitfeldt: I would change this to `expect_snapshot()`

CharlotteZZh:
> do we have any issues that we can link to for this PR or was it just random things you found while you were working here?

I use Codecov and sort by 'missed' in the R folder, then work through it in that order.

EmilHvitfeldt (approved):

PR #1483, opened 2026-07-28
Add tests for the bag

CharlotteZZh:
Unit tests for bagControl and bag()'s argument validation, plus integration tests that fit an ldaBag ensemble through the default and formula interfaces and exercise predict/print/summary (all snapshotted).


topepo (changes requested):
The dependencies for the other bagged models (e.g., nnet, pls, etc) are in Suggests, so you can also add those too.

PR #1484, opened 2026-07-28
Add tests for avNNet

CharlotteZZh:
Integration tests for avNNet across the default, formula, classification and regression paths, exercising all predict types, the print method (snapshotted), and the predict class-check error. Raises R/avNNet.R coverage from 0% to ~84%.


topepo (changes requested):
Can you also test that it fits and predicts when:

- there is a single predictor in the model
- a single row is being predicted.

PR #1493, opened 2026-08-10
Add tests for knn3 and knnreg

CharlotteZZh:
One test file covers the k-nearest-neighbour classification (knn3) and regression (knnreg) wrappers, which share the same formula/matrix/data.frame interfaces, print, predict and low-level *Train engines.


topepo (changes requested):
Looks good but please add tests for a single predictor as well as predicting a single row

PR #1496, opened 2026-08-11
Add tests for icr and fix its broken formula interface

CharlotteZZh:
Cover independent component regression across the default and formula interfaces, single-predictor and single-row prediction, the numeric-outcome error and print. Testing the formula interface surfaced a bug: icr.formula forwarded an unused weights argument (and an undefined thresh) into fastICA, which errored; it now forwards ... cleanly.

### DIFF
--- a/R/icr.R
+++ b/R/icr.R
@@ -72,7 +72,7 @@ icr.formula <- function(
   }
   y <- model.response(m)
 
-  res <- icr.default(x, y, weights = w, thresh = thresh, ...)
+  res <- icr.default(x, y, ...)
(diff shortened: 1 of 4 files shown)


R/icr.R
```diff
@@ -72,7 +72,7 @@ icr.formula <- function(
   }
   y <- model.response(m)
 
-  res <- icr.default(x, y, weights = w, thresh = thresh, ...)
+  res <- icr.default(x, y, ...)
```
topepo, line 75: nice catch

topepo (approved):

